PR #2, opened 2026-05-06
Add multi-engine AI support with OpenCode integration

carlesarnal:
## Summary

Introduces a pluggable, multi-engine AI architecture. Axiom can now use multiple AI engines simultaneously — each action type can specify which engine to use (e.g. OpenCode for some tasks, Claude Code for others), with a global default as fallback.

## Architecture

```
axiom.ai-engine=claude-code          (global default)
Action Type "analyze" → engine: null  (uses default → Claude Code)
Action Type "implement" → engine: opencode  (uses OpenCode)
Action Type "review" → engine: null   (uses default → Claude Code)
```

### `engine/spi/` — Engine abstraction layer
- **`AiEngine`** — Core SPI: `prompt()`, `promptWithSchema()`, `healthCheck()`
- **`AiEngineProvider`** — CDI discovery interface
- **`AiEngineRegistry`** — Holds all available engines, resolves by type with fallback
- **`AiEngineProducer`** — CDI producer for the global default engine
- **`AiEngineConfig`**, **`AiEngineResult`**, **`AiEngineMcpManager`**

### `engine/opencode/` — OpenCode engine
- **`OpenCodeEngine`** — HTTP API client to \`opencode serve\`
- **`OpenCodeActor`** — Actor SPI with session tracking + cancellation
- **`OpenCodeServerManager`** — Server lifecycle with health monitoring
- **`OpenCodePermissionMapper`** — Axiom tools → OpenCode permissions
- **`OpenCodeMcpManager`** — Dynamic MCP registration via \`POST /mcp\`

### Claude Code engine (in `actors/claude-code/`)
- **`ClaudeCodeEngine`** — Wraps existing subprocess logic behind the SPI

## Per-action-type engine selection
- `ActionTypeEntity.engine` — nullable column, overrides global default
- `AiEngineRegistry` — resolves engine per action type, falls back to default
- `TaskExecutionService` — uses registry to pick the right engine and actor
- `GET /system/engines` — lists available engine types for the UI

## Refactored consumers (engine-agnostic)
- **`ManagerService`**, **`ScriptAiService`**, **`ToolAiService`**, **`ReportExecutionService`** — use \`AiEngine\` SPI
- **`TaskExecutionService`** — resolves engine per action type via \`AiEngineRegistry\`
- **`StartupCheckService`** — delegates to \`AiEngine.healthCheck()\`

## UI changes
- **Engine indicator** in masthead — clickable badge showing active engine
- **AI Engine settings page** — health checks, available models grouped by provider
- **Engine selector** on Action Type detail page — dropdown (shown when >1 engine available)
- **Provider-grouped model picker** — models grouped under Anthropic/OpenAI headings

## REST API
- \`GET /system/config\` — \`engine\` field shows active engine type
- \`GET /system/models\` — dynamic model list per engine
- \`GET /system/engines\` — lists available engines
- \`ActionType\` schema — new \`engine\` field

## Tests
105 total (95 passing, 10 skipped pre-existing Claude CLI integration tests)

## How to use

```properties
# Global default engine
axiom.ai-engine=claude-code

# Per-action-type override (in the UI or via REST API)
# Set engine: "opencode" on specific action types
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ long-lived projects around them, and delegates work to human and AI actors.
 
 - Watches GitHub/Jira repositories for issue activity (webhooks or polling)
 - An AI Manager triages events and decides what actions to take
-- Tasks are assigned to Actors (Claude Code AI agents or humans)
+- Tasks are assigned to Actors (AI agents or humans)
 - Projects track the full lifecycle of work tied to each issue
 - A UI provides visibility into projects, tasks, activity, and configuration
 
@@ -18,16 +18,28 @@ long-lived projects around them, and delegates work to human and AI actors.
 | Backend | Java 25 / Quarkus 3.33 LTS |
 | Frontend | TypeScript / React / PatternFly |
 | Database | H2 (dev) / PostgreSQL (prod) |
-| AI | Claude Code CLI |
+| AI | Pluggable: Claude Code CLI or OpenCode (configurable) |
 | API | Contract-first OpenAPI + apicurio-codegen |
+
+## AI Engine
+
+Axiom's AI engine is pluggable. Set `axiom.ai-engine` in `application.properties` to
+select which engine to use:
+
+| Engine | Config Value | Binary | Install |
+|--------|-------------|--------|---------|
+| Claude Code | `claude-code` (default) | `claude` | `npm install -g @anthropic-ai/claude-code` |
+| OpenCode | `opencode` | `opencode` | `curl -fsSL https://opencode.ai/install \| bash` |
+
+Only the selected engine's binary needs to be installed. Both can coexist for testing.
 
 ## Prerequisites
 
 - Java 25+
 - Maven 3.9+
-- Node.js 20+ (for UI)
-- Claude Code CLI (for AI features)
-- `ANTHROPIC_API_KEY` environment variable
+- Node.js 20+ (for UI and MCP tool server)
+- One of the AI engine CLIs (see table above)
+- API key for your LLM provider (e.g. `ANTHROPIC_API_KEY`)
 
 ## Build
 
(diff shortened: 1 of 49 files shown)


carlesarnal:
@EricWittmann what do you think about this? This is the starting point to allow using other ai engines, mainly opencode.

EricWittmann:
> @EricWittmann what do you think about this? This is the starting point to allow using other ai engines, mainly opencode.

What do I think??  I think:  "yes please"  :)

EricWittmann:
Do we want to support multiple engines enabled simultaneously at some point?

carlesarnal:
> Do we want to support multiple engines enabled simultaneously at some point?

I guess it depends on the meaning we give to the "ai engine" concept. If it's multiple models, with what's here you can switch (not use them simultaneously). For tools, similar thing. Does it really make sense to allow different "tools" (claude code, opencode) at the same time? I see a reason for models, since one can be better at a certain task that another, but I'm not sure it's that valuable for tools.

EricWittmann:
> > Do we want to support multiple engines enabled simultaneously at some point?
> 
> I guess it depends on the meaning we give to the "ai engine" concept. If it's multiple models, with what's here you can switch (not use them simultaneously). For tools, similar thing. Does it really make sense to allow different "tools" (claude code, opencode) at the same time? I see a reason for models, since one can be better at a certain task that another, but I'm not sure it's that valuable for tools.

My only thinking would be to use Bob for certain things and Claude for others... for obvious reasons.  But I suppose you could just as easily run two instances of Axiom.

carlesarnal:
> > > Do we want to support multiple engines enabled simultaneously at some point?
> > 
> > 
> > I guess it depends on the meaning we give to the "ai engine" concept. If it's multiple models, with what's here you can switch (not use them simultaneously). For tools, similar thing. Does it really make sense to allow different "tools" (claude code, opencode) at the same time? I see a reason for models, since one can be better at a certain task that another, but I'm not sure it's that valuable for tools.
> 
> My only thinking would be to use Bob for certain things and Claude for others... for obvious reasons. But I suppose you could just as easily run two instances of Axiom.

Well, if you could give opencode or claude code access to IBM models there would be no need for that but yes, I see why you want that. I think it should be relatively easy to allow it.

PR #6, opened 2026-05-15
Fix #5: add missing database sequences for PanacheEntity tables

orpiske:
## Summary

- Adds Flyway migration `V17__create_missing_sequences.sql` with the 15 missing database sequences required by PanacheEntity ID generation
- Production startup fails because dev mode uses `drop-and-create` (Hibernate auto-creates sequences) while production uses Flyway only, which never created them

## Test plan

- [x] All 77 existing tests pass
- [ ] Run with `java -jar quarkus-run.jar` against a persistent database and verify inserts succeed

PR #8, opened 2026-05-15
Fix #7: make Claude Code executable path configurable

### LINKED ISSUE #7: ClaudeCodeSubprocess fails to find claude binary when not on PATH
## Description

`ClaudeCodeSubprocess` fails with `IOException: Cannot run program "claude"` when the `claude` CLI is not on the Java process's PATH. This happens because `ClaudeCodeCommandBuilder` hardcodes `"claude"` as the executable name.

On macOS with zsh, the shell profile (`.zshrc`) typically adds `claude` to the PATH, but the Java process does not inherit that environment, causing the subprocess to fail:

```
java.io.IOException: Cannot run program "claude" (in directory "/Users/.../.axiom/workspaces/project-1"): Exec failed, error: 2 (No such file or directory)
    at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1112)
    at io.apicurio.axiom.actors.claudecode.ClaudeCodeSubprocess.doExecute(ClaudeCodeSubprocess.java:121)
```

## Proposed Fix

Add a configurable property `axiom.claude-code.executable` (defaulting to `"claude"`) so users can specify the full path to the binary when it is not on the system PATH.

orpiske:
## Summary

- Add `axiom.claude-code.executable` config property (defaults to `"claude"`) to allow specifying the full path to the Claude CLI binary
- On macOS with zsh, the Java process does not inherit shell profile PATH additions, so `ProcessBuilder` fails to locate `claude`

## Changes

- `ClaudeCodeCommandBuilder`: add `executable` field and setter, use it instead of hardcoded `"claude"`
- `ClaudeCodeActor`: inject the new config property and pass it to the builder
- `ClaudeCodeCommandBuilderTest`: add test for custom executable path

## Test plan

- [x] Existing tests pass (34 tests, 0 failures)
- [x] New `testCustomExecutable` test verifies custom path is used as first command element
- [x] Full reactor build passes


orpiske:
This also has the fix in the pr #6, but just out of convenience for my own testing. When/if approved, I'll make sure to rebase and clean it up for inclusion.

orpiske:
On a second inspection, I actually think this was because it could not find the **project** folder. Not `claude`. I'll leave it open to collect feedback, but I think it would be ok to skip this one.

EricWittmann:
Oh man, I'm sorry @orpiske - I saw the GH issue but NOT the attached PR you opened.  :(

Then I went ahead and implemented the suggested fix (which was reasonable and simple).  My bad.

PR #24, opened 2026-06-18
Fix Create Report Definition modal failing due to empty promptTemplate (#20)

### LINKED ISSUE #20: Create Report Definition modal fails due to promptTemplate validation
## Bug

The **Create Report Definition** modal dialog no longer works. Clicking "Create" results in a
422 error from the backend.

## Root Cause

The modal sends a minimal payload with an empty `promptTemplate`:

```json
{ "name": "...", "schedule": "daily", "timeWindow": "last-24h", "promptTemplate": "", "enabled": false }
```

The backend `ReportDefinitionValidator.validatePromptTemplate()` treats an empty/blank prompt
template as a validation error:

```java
if (prompt == null || prompt.isBlank()) {
    messages.add(error("promptTemplate", "Prompt template is required."));
```

The `createReportDefinition` endpoint calls `validateOrThrow()`, which returns HTTP 422 when
there are any validation errors. Since the modal intentionally omits the prompt template (the
user is expected to fill it in on the detail page), creation always fails.

## Affected Files

- `ui/src/pages/ReportDefinitionsPage.tsx` — modal sends `promptTemplate: ""`

## Fix

Update the modal's `handleCreate` function in `ReportDefinitionsPage.tsx` to send a default
"hello world" style prompt template instead of an empty string, so that backend validation
passes.

EricWittmann:
## Summary

- Send a default prompt template instead of an empty string when creating a new report definition
- The empty `promptTemplate` was rejected by backend validation (`ReportDefinitionValidator`), causing a 422 error

## Related Issue

Fixes #20

## Test Plan

- [ ] Open the Report Definitions page
- [ ] Click "Create" to open the modal, enter a name and schedule, then click Create
- [ ] Verify the report definition is created successfully (no 422 error)
- [ ] Verify you are navigated to the new report definition's detail page
- [ ] Verify the default prompt template text appears on the detail page

### DIFF
--- a/ui/src/pages/ReportDefinitionsPage.tsx
+++ b/ui/src/pages/ReportDefinitionsPage.tsx
@@ -65,7 +65,7 @@ export function ReportDefinitionsPage() {
             name: newName,
             schedule: newSchedule,
             timeWindow: "last-24h",
-            promptTemplate: "",
+            promptTemplate: "Summarize recent activity across all repositories.",
             enabled: false,
         })
             .then((def) => {


EricWittmann (commented):
## Code Review Summary

This PR fixes a bug where clicking "Create" in the Create Report Definition modal resulted in a 422 validation error because the `promptTemplate` field was sent as an empty string, which the backend `ReportDefinitionValidator` rejects. The fix replaces the empty string with a sensible default prompt: `"Summarize recent activity across all repositories."`.

---

### Findings

**No significant issues found.** This is a clean, minimal, one-line fix that directly addresses the reported bug.

A couple of minor observations (not blocking):

#### 💬 Nitpick — Inline magic string

**File:** `ui/src/pages/ReportDefinitionsPage.tsx` (line 68)

The default prompt template is an inline string literal. For improved maintainability, it could be extracted to a named constant (e.g., `DEFAULT_PROMPT_TEMPLATE`) — especially if this default might be referenced elsewhere in the future. That said, it's currently used in a single place and the string is self-descriptive, so this is purely a style preference.

#### ✅ Consistency check — Detail page is fine

The `ReportDefinitionDetailPage` initializes its form state with `promptTemplate: ""`, but this is only a transient React state default that gets immediately replaced by the fetched data from the API. No issue there.

---

### Overall Assessment

✅ **Looks good to merge.** The change is correct, minimal, and directly fixes the reported issue. The default prompt is a reasonable starting point that users can customize on the detail page after creation.

PR #28, opened 2026-06-22
Color-code labels based on text value (#27)

### LINKED ISSUE #27: Color-code labels based on text value instead of always purple
## Feature

User-defined labels on Reports, Tools, and Projects are always rendered in purple. This makes
it hard to visually distinguish labels at a glance.

## Proposed Change

Deterministically assign a PatternFly Label color based on the label's text value using a hash
function, so that:

- The same label text always gets the same color
- Different label texts are distributed across all 9 available PatternFly Label colors:
  blue, teal, green, orange, purple, red, orangered, grey, yellow

## Implementation

Create a new `ColoredLabel` React component that wraps PatternFly's `Label` and automatically
computes the color from the children text. Replace all user-defined label rendering (list pages
and shared components) with this new component. Status/type labels that use intentional semantic
colors are not affected.

EricWittmann:
## Summary

- Add a new `ColoredLabel` component that wraps PatternFly's `Label` and deterministically
  assigns a color based on the label text using a hash function
- Replace all user-defined label rendering (5 files) with `ColoredLabel`, distributing labels
  across all 9 PatternFly colors instead of always purple
- Status/type labels with intentional semantic colors are unchanged

## Related Issue

Fixes #27

## Test Plan

- [ ] Navigate to Reports, Tools, and Projects list pages — labels should display in varied colors
- [ ] Verify the same label text always renders in the same color across pages
- [ ] Verify label click-to-filter still works on list pages
- [ ] Open a detail page (Report, Tool, or Project) and verify label editing (add/remove) works
- [ ] Confirm status labels (e.g. Completed, Failed) retain their original semantic colors

### DIFF
--- a/ui/src/components/ColoredLabel.tsx
+++ b/ui/src/components/ColoredLabel.tsx
@@ -1,0 +1,27 @@
+import { Label, type LabelProps } from "@patternfly/react-core";
+
+const LABEL_COLORS: LabelProps["color"][] = [
+    "blue", "teal", "green", "orange", "purple",
+    "red", "orangered", "grey", "yellow",
+];
+
+/**
+ * Deterministically maps a label string to a PatternFly label color
+ * so that the same text always produces the same color.
+ */
+export function labelColor(text: string): LabelProps["color"] {
+    let hash = 0;
+    for (let i = 0; i < text.length; i++) {
+        hash = (hash * 31 + text.charCodeAt(i)) | 0;
+    }
+    return LABEL_COLORS[Math.abs(hash) % LABEL_COLORS.length];
+}
+
+/**
+ * A PatternFly Label whose color is automatically determined by
+ * its text content. Accepts all standard Label props except `color`.
+ */
+export function ColoredLabel({ children, ...rest }: Omit<LabelProps, "color">) {
+    const text = typeof children === "string" ? children : String(children);
+    return <Label color={labelColor(text)} {...rest}>{children}</Label>;
+}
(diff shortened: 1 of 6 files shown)


EricWittmann (commented):
## Code Review — Color-code labels based on text value (#27)

This PR introduces a new `ColoredLabel` component that deterministically maps label text to one of nine PatternFly colors via a hash function, replacing the previously hardcoded `color="purple"` across five files. Status labels with intentional semantic colors are correctly left unchanged.

### Findings

#### 🟡 Nitpick

**`ui/src/components/LabelDisplay.tsx` — Stale JSDoc comment**

The component's JSDoc still reads *"Read-only display of labels as **blue** chips…"* (line 5), but labels now render in varied colors via `ColoredLabel`. Consider updating to something like *"Read-only display of labels as colored chips…"* to keep the docs accurate.

### Overall Assessment

✅ **Looks good to merge.**

The PR is clean and well-scoped:

- The hash function (`hash * 31 + charCode`) is a standard, well-understood approach (similar to Java's `String.hashCode()`). The `| 0` bitwise OR correctly keeps the accumulator within 32-bit integer range to prevent floating-point drift, and `Math.abs` on a JS number (64-bit float) is safe even for the edge-case value of `-2147483648`.
- Exporting the `labelColor` function separately from the component is a nice touch — makes it independently testable.
- The `Omit<LabelProps, "color">` type correctly prevents callers from overriding the auto-assigned color while preserving all other `Label` props (including `onClose` used in `LabelInput`).
- The `String(children)` fallback in `ColoredLabel` is a reasonable defensive measure, even though all current call sites pass string children.
- All five consumer files are updated consistently, and status labels (project status, report status) properly continue using the standard `Label` with explicit `STATUS_COLORS`.

No correctness, security, or performance concerns. Nice improvement! 🎨

PR #42, opened 2026-06-22
Fix AI Assistant message input not shrinking after submit (#37)

### LINKED ISSUE #37: AI Assistant message input does not shrink back after submitting
## Bug

The AI Assistant's "Type a message" text area auto-expands as the user types more text,
which is correct. However, after the user submits the message, the text area remains at
its expanded size instead of reverting to its original compact height.

## Steps to Reproduce

1. Open the AI Assistant
2. Type a multi-line message (long enough that the text area grows)
3. Submit the message (press Enter or click Send)
4. Observe that the text area stays at the expanded height despite being empty

## Expected Behavior

After submission, the text area should reset to its initial compact size (single-line
height), matching its appearance when no text has been entered.

EricWittmann:
## Summary
- PatternFly's `TextArea` `autoResize` only recalculates height on user-generated change events,
  not on programmatic value changes
- After `handleSend` clears the value with `setValue("")`, the parent span retains its expanded
  inline height
- Added a ref to the textarea and reset the parent element's inline height after clearing, so the
  textarea collapses back to its natural `rows={1}` height

## Related Issue
Fixes #37

## Test Plan
- [ ] Open the AI Assistant
- [ ] Type a multi-line message that causes the textarea to expand
- [ ] Submit the message (Enter or click Send)
- [ ] Verify the textarea shrinks back to its initial compact height
- [ ] Type another multi-line message and verify auto-expand still works

### DIFF
--- a/ui/src/components/assistant/AssistantMessageInput.tsx
+++ b/ui/src/components/assistant/AssistantMessageInput.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback } from "react";
+import { useState, useCallback, useRef } from "react";
 import { TextArea, Button, Flex, FlexItem } from "@patternfly/react-core";
 import PaperPlaneIcon from "@patternfly/react-icons/dist/esm/icons/paper-plane-icon";
 
@@ -9,12 +9,16 @@ interface AssistantMessageInputProps {
 
 export function AssistantMessageInput({ onSend, disabled }: AssistantMessageInputProps) {
     const [value, setValue] = useState("");
+    const textAreaRef = useRef<HTMLTextAreaElement>(null);
 
     const handleSend = useCallback(() => {
         const trimmed = value.trim();
         if (!trimmed) return;
         onSend(trimmed);
         setValue("");
+        setTimeout(() => {
+            textAreaRef.current?.parentElement?.style.removeProperty("height");
+        }, 0);
     }, [value, onSend]);
 
     const handleKeyDown = (e: React.KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review Summary

This PR fixes a UI bug where the AI Assistant's textarea doesn't shrink back to its initial height after submitting a message. The root cause is that PatternFly's `TextArea` `autoResize` only recalculates height on user-generated `change` events, not on programmatic `setValue("")` calls. The fix adds a ref to the textarea and manually removes the inline `height` style from its parent element after clearing the value.

---

### Findings

#### 💡 Suggestion

**`AssistantMessageInput.tsx` — DOM structure coupling with `parentElement`**

```ts
textAreaRef.current?.parentElement?.style.removeProperty("height");
```

This relies on PatternFly's internal DOM structure — specifically that the `<textarea>` is wrapped in a parent element (a `<span>`) that receives the inline `height` style from `autoResize`. If PatternFly changes this structure in a future version (e.g., adding an intermediate wrapper or moving the style to a different element), this fix would silently stop working.

This is an acceptable tradeoff for a targeted workaround, but it would be helpful to add a brief comment explaining *why* this works and what it depends on, so that future maintainers (or anyone upgrading PatternFly) know to verify it:

```ts
// PatternFly autoResize sets an inline height on the textarea's parent <span>.
// Programmatic value changes don't trigger a recalc, so we manually clear it.
textAreaRef.current?.parentElement?.style.removeProperty("height");
```

#### 🔍 Nitpick

**`AssistantMessageInput.tsx` — `setTimeout` with 0ms delay**

The `setTimeout(..., 0)` is necessary to defer the DOM manipulation until after React's state flush from `setValue("")`. This is a well-known pattern and works correctly here. Just noting that this is inherently timing-dependent — if React batches updates differently in a future version, you might need to revisit. A brief inline comment would help here too:

```ts
// Defer to next tick so React flushes setValue("") before we reset height
setTimeout(() => { ... }, 0);
```

---

### Overall Assessment

**✅ Looks good to merge.**

This is a clean, minimal, and well-scoped fix. The approach is a reasonable workaround for a known PatternFly `autoResize` limitation. The optional chaining provides appropriate null-safety, and the `useCallback` dependency array is correct. The only suggestion is to add a couple of explanatory comments to make the workaround self-documenting for future maintenance.

PR #43, opened 2026-06-23
Fix TransactionRequiredException in ManagerService during pipeline event processing (#31)

### LINKED ISSUE #31: TransactionRequiredException in ManagerService during pipeline event processing
## Bug

The pipeline orchestrator's scheduled event processing fails with a
`TransactionRequiredException` when the Manager evaluates an event. The error occurs in
`ManagerService.recordAiUsage()` and `ManagerService.logManagerActivity()`, both called from
`ManagerService.evaluate()`.

## Stack Trace Summary

```
ERROR [io.ap.ax.ma.ManagerService] Manager evaluation failed for event 1751:
  jakarta.persistence.TransactionRequiredException:
    Transaction is not active, consider adding @Transactional to your method...

  at ManagerService.recordAiUsage(ManagerService.java:242)
  at ManagerService.evaluate(ManagerService.java:114)
  at PipelineOrchestrator.processEvent(PipelineOrchestrator.java:96)
  at PipelineOrchestrator.processNextEvent(PipelineOrchestrator.java:69)
```

A second error follows immediately for `logManagerActivity` (line 227), called later in
`evaluate()` (line 149). The outer transaction from `PipelineOrchestrator.processEvent()`
has already been marked for rollback after the first failure, so all subsequent persistence
operations in the same transaction also fail.

The final exception is a `RollbackException: ARJUNA016102: The transaction is not active!`
when the `@Transactional` interceptor on `processEvent()` tries to commit.

## Root Cause

`ManagerService.recordAiUsage()` and `logManagerActivity()` both call `entity.persist()`
and appear to have `@Transactional(REQUIRED)` semantics. When called from within the
existing transaction started by `PipelineOrchestrator.processEvent()`, a failure in
`recordAiUsage()` marks the outer transaction for rollback. The subsequent call to
`logManagerActivity()` then fails because the transaction is no longer active.

## Affected Files

- `manager/src/main/java/io/apitomy/axiom/manager/ManagerService.java` — lines 114, 149,
  227, 242
- `app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java` — lines 69, 96

## Steps to Reproduce

1. Start Axiom with `./dev.sh --persist` against a database with pending events
2. Observe the scheduled `processNextEvent` trigger pick up an event
3. The Manager evaluation completes (AI call succeeds) but persisting the AI usage record
   fails with `TransactionRequiredException`

## Environment

- Quarkus 3.33.1
- H2 file-based database (persist mode)
- Java 25

EricWittmann:
## Summary

- Add `@Transactional` to `PipelineOrchestrator.processNextEvent()` so the scheduler-invoked
  CDI proxy starts a transaction for the entire event processing pipeline. Previously,
  `processEvent()` had `@Transactional` but was called via self-invocation from
  `processNextEvent()`, bypassing the CDI interceptor entirely.
- Replace `@Transactional` on `ManagerService.recordAiUsage()` and `logManagerActivity()`
  with `QuarkusTransaction.requiringNew()` programmatic transactions. These methods were
  also called via self-invocation from `evaluate()`, making the annotations dead code.
  Using `REQUIRES_NEW` ensures AI usage and activity log entries are committed independently
  of the outer transaction.

## Related Issue

Fixes #31

## Test Plan

- [ ] Build passes (`./build.sh`)
- [ ] Start Axiom with `./dev.sh --persist` against a database with pending events
- [ ] Observe scheduled `processNextEvent` picks up events without `TransactionRequiredException`
- [ ] Verify AI usage records are persisted in the database after Manager evaluation
- [ ] Verify activity log entries are created for Manager decisions

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -70,6 +70,8 @@ public class PipelineOrchestrator {
         processEvent(queueEntry);
     }
 
+    // @Transactional is redundant here (runs inside processNextEvent's transaction via
+    // self-invocation) but retained so it takes effect if called through the CDI proxy.
     @Transactional
     EventQueueEntity findNextPendingEvent() {
         EventQueueEntity entry = EventQueueEntity.find(
@@ -80,6 +82,8 @@ public class PipelineOrchestrator {
         return entry;
     }
 
+    // @Transactional is redundant here (runs inside processNextEvent's transaction via
+    // self-invocation) but retained so it takes effect if called through the CDI proxy.
     @Transactional
     void processEvent(EventQueueEntity queueEntry) {
         EventEntity event = EventEntity.findById(queueEntry.eventId);
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -58,6 +58,7 @@ public class PipelineOrchestrator {
      * Polls the event queue every 5 seconds and processes one pending event at a time.
      * Events are processed sequentially to avoid race conditions in the Manager.
      */
+    @Transactional
     @Scheduled(every = "${axiom.pipeline.poll-interval:5s}",
                concurrentExecution = Scheduled.ConcurrentExecution.SKIP)
     void processNextEvent() {
--- a/manager/src/main/java/io/apitomy/axiom/manager/ManagerService.java
+++ b/manager/src/main/java/io/apitomy/axiom/manager/ManagerService.java
@@ -15,7 +15,7 @@ import io.apitomy.axiom.engine.spi.AiEngineConfig;
 import io.apitomy.axiom.engine.spi.AiEngineResult;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
-import jakarta.transaction.Transactional;
+import io.quarkus.narayana.jta.QuarkusTransaction;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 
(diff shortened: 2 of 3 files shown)


EricWittmann (commented):
## Code Review

This PR fixes a well-known CDI self-invocation issue where `@Transactional` annotations on `processEvent()`, `recordAiUsage()`, and `logManagerActivity()` were silently bypassed because those methods were invoked on `this` rather than through the CDI proxy. The fix correctly adds `@Transactional` to the scheduler entry point (`processNextEvent()`) and switches the `ManagerService` methods to programmatic `QuarkusTransaction.requiringNew()` transactions.

---

### ⚠️ Warning

**Long-running transaction in `processNextEvent()` — `PipelineOrchestrator.java`**

Adding `@Transactional` to `processNextEvent()` means the entire pipeline — including the AI engine call (`aiEngine.promptWithSchema().join()`) — runs inside a single JTA transaction. The AI call can block for up to `axiom.manager.timeout-seconds` (default 120s), holding a database connection from the pool for the full duration.

While `ConcurrentExecution.SKIP` prevents concurrent scheduler invocations, a 120-second connection hold is still significant. If other parts of the application also need connections during that window, this could lead to pool contention. Worth keeping an eye on in production.

A future improvement could split the pipeline into phases: a short transaction to dequeue and mark the event as "processing", then the AI call outside any transaction, then another transaction to persist decisions. However, this is a larger refactor and not necessarily required for this fix.

---

### 💡 Suggestions

**1. Dead `@Transactional` on `findNextPendingEvent()` and `processEvent()` — `PipelineOrchestrator.java:73-84`**

Both `findNextPendingEvent()` and `processEvent()` are called via self-invocation from `processNextEvent()`, so their `@Transactional` annotations are also bypassed by CDI — they rely entirely on the outer transaction from `processNextEvent()`. The annotations are harmless but misleading: they imply these methods manage their own transaction boundaries when they don't.

Consider either removing the annotations (to avoid confusion) or adding a brief comment explaining that they're retained for documentation/future use.

**2. Consider making `recordAiUsage()` non-fatal — `ManagerService.java:114-115`**

If `recordAiUsage()` throws (e.g., a constraint violation in the `REQUIRES_NEW` transaction), the exception propagates into the `catch` block in `evaluate()`, which means the successfully-returned AI decisions are discarded and the event is treated as a failure. Since AI usage recording is an auxiliary/audit concern, consider wrapping it in a try-catch so a recording failure doesn't prevent the actual decisions from being processed:

```java
try {
    recordAiUsage(event.id, project != null ? project.id : null,
            result.costUsd(), result.inputTokens(), result.outputTokens());
} catch (Exception e) {
    LOG.warnf(e, "Failed to record AI usage for event %d", event.id);
}
```

---

### ✅ What looks good

- **Root cause analysis is correct.** CDI self-invocation bypassing interceptors is a classic pitfall, and the PR description explains the issue clearly.
- **`REQUIRES_NEW` semantics for audit data** is a sound design choice — AI usage and activity log entries should be committed independently so they survive outer transaction rollbacks.
- **The `QuarkusTransaction.requiringNew()` approach** avoids the self-invocation trap entirely since it's programmatic, not annotation-driven.

---

### Overall

**Looks good to merge** with the minor suggestions above. The core fix is correct and well-motivated. The long-running transaction note is something to be aware of but not a blocker for this PR.

PR #56, opened 2026-06-24
Tracing: Tool call tracing via MCP server instrumentation (#48)

### LINKED ISSUE #48: Tracing: Tool call tracing via MCP server instrumentation
## Parent Epic
Part of #36 — End-to-end activity tracing and execution lineage.
Depends on #45 (data model), #46 (backend instrumentation, which passes trace
context into MCP config as env vars), and #49 (REST API, which defines the OpenAPI
contract for the callback endpoint — required for contract-first code generation).

## Summary
Instrument the generated MCP servers so that every custom tool and SDK tool call
creates a trace node with full structured input/output. Axiom controls MCP server
generation, so we can wrap tool handlers with trace callbacks at the generation layer.

### MCP Config Generation (McpConfigGenerator.java)

Issues #46 and #47 handle injecting the trace env vars into the MCP config:
```java
env.put("AXIOM_TRACE_ID", traceCtx.traceId().toString());
env.put("AXIOM_PARENT_NODE_ID", String.valueOf(parentNodeId));
```

These env vars are already inherited by the MCP server subprocesses because
`McpConfigGenerator` includes them in the `env` block of each MCP server entry.

**No changes needed to McpConfigGenerator in this issue** — the env var injection is
handled by #46/#47. This issue focuses on the MCP server-side instrumentation and the
callback endpoint.

### SDK Server Instrumentation (sdk-server.js)

The SDK server (`templates/axiom-mcp-server/sdk-server.js`) handles tool calls via
a `CallToolRequestSchema` handler. Wrap the tool execution with trace callbacks:

```javascript
// Read trace context from environment
const TRACE_ID = process.env.AXIOM_TRACE_ID;
const PARENT_NODE_ID = process.env.AXIOM_PARENT_NODE_ID;
const AXIOM_API_URL = process.env.AXIOM_API_URL;

server.setRequestHandler(CallToolRequestSchema, async (request) => {
    const tool = SDK_TOOLS.find(t => t.name === request.params.name);
    if (!tool) throw new Error(`Unknown tool: ${request.params.name}`);

    const toolInput = JSON.stringify(request.params.arguments || {});
    let traceNodeId = null;

    // Before: create trace node (if tracing is active)
    if (TRACE_ID && PARENT_NODE_ID && AXIOM_API_URL) {
        try {
            const resp = await fetch(`${AXIOM_API_URL}/traces/tool-calls`, {
                method: "POST",
                headers: { "Content-Type": "application/json" },
                body: JSON.stringify({
                    traceId: TRACE_ID,
                    parentNodeId: parseInt(PARENT_NODE_ID),
                    toolName: request.params.name,
                    toolInput: toolInput,
                    status: "in-progress"
                })
            });
            if (resp.ok) {
                const data = await resp.json();
                traceNodeId = data.nodeId;
            }
        } catch (e) {
            // Non-fatal: don't break tool execution if tracing fails
            console.error("Trace callback failed:", e.message);
        }
    }

    // Execute tool
    let result, status = "success";
    const startTime = Date.now();
    try {
        result = await tool.handler(request.params.arguments || {});
    } catch (e) {
        status = "failure";
        result = `Error: ${e.message}`;
    }
    const durationMs = Date.now() - startTime;

    // After: complete trace node
    if (traceNodeId && AXIOM_API_URL) {
        try {
            await fetch(`${AXIOM_API_URL}/traces/tool-calls/${traceNodeId}`, {
                method: "PUT",
                headers: { "Content-Type": "application/json" },
                body: JSON.stringify({
                    toolOutput: typeof result === "string"
                        ? result.substring(0, 10000)   // Truncate large outputs
                        : JSON.stringify(result).substring(0, 10000),
                    status: status,
                    durationMs: durationMs
                })
            });
        } catch (e) {
            console.error("Trace completion callback failed:", e.message);
        }
    }

    if (status === "failure") throw new Error(result);
    return { content: [{ type: "text", text: result || "OK" }] };
});
```

### Tools Server Instrumentation (tools-server.js)

The tools server (`templates/axiom-mcp-server/tools-server.js`) executes bash script
tools via `execSync()`. Apply the same wrapping pattern:

```javascript
server.setRequestHandler(CallToolRequestSchema, async (request) => {
    const tool = SCRIPT_TOOLS.find(t => t.name === request.params.name);
    const toolInput = JSON.stringify(request.params.arguments || {});
    let traceNodeId = null;

    // Before: create trace node
    if (TRACE_ID && PARENT_NODE_ID && AXIOM_API_URL) {
        // Same fetch pattern as sdk-server.js
    }

    // Execute tool (existing execSync logic)
    let result, status = "success";
    const startTime = Date.now();
    try {
        // ... existing script execution logic ...
    } catch (e) {
        status = "failure";
        result = e.stderr || e.message;
    }
    const durationMs = Date.now() - startTime;

    // After: complete trace node
    if (traceNodeId && AXIOM_API_URL) {
        // Same fetch pattern as sdk-server.js
    }

    return { content: [{ type: "text", text: result }] };
});
```

### Shared Trace Helper Module

To avoid duplicating trace callback code between sdk-server.js and tools-server.js,
create a shared helper:

```javascript
// templates/axiom-mcp-server/trace-helper.js
const TRACE_ID = process.env.AXIOM_TRACE_ID;
const PARENT_NODE_ID = process.env.AXIOM_PARENT_NODE_ID;
const AXIOM_API_URL = process.env.AXIOM_API_URL;

const tracingEnabled = !!(TRACE_ID && PARENT_NODE_ID && AXIOM_API_URL);

async function startToolTrace(toolName, toolInput) {
    if (!tracingEnabled) return null;
    try {
        const resp = await fetch(`${AXIOM_API_URL}/traces/tool-calls`, { ... });
        if (resp.ok) return (await resp.json()).nodeId;
    } catch (e) { console.error("Trace start failed:", e.message); }
    return null;
}

async function completeToolTrace(nodeId, toolOutput, status, durationMs) {
    if (!nodeId) return;
    try {
        await fetch(`${AXIOM_API_URL}/traces/tool-calls/${nodeId}`, { ... });
    } catch (e) { console.error("Trace complete failed:", e.message); }
}

module.exports = { startToolTrace, completeToolTrace, tracingEnabled };
```

### Trace Callback REST Endpoint (Java)

New endpoint in the Axiom REST API for receiving tool call trace data:

**POST /api/v1/traces/tool-calls** — Create tool trace node
```json
Request:
{
    "traceId": "550e8400-e29b-41d4-a716-446655440000",
    "parentNodeId": 42,
    "toolName": "mcp__axiom-sdk__axiom_list_projects",
    "toolInput": "{\"status\": \"active\"}",
    "status": "in-progress"
}

Response (201):
{
    "nodeId": 57,
    "toolExecutionId": 12
}
```

This endpoint atomically creates:
1. `ToolExecutionEntity` with traceId, toolName, toolInput, status
2. `TraceNodeEntity` with traceId, parentNodeId, nodeType="tool-call",
   entityType="tool-execution", entityId=toolExecutionEntity.id

**PUT /api/v1/traces/tool-calls/{nodeId}** — Complete tool trace node
```json
Request:
{
    "toolOutput": "{ ... }",
    "status": "success",
    "durationMs": 1234
}
```

This endpoint updates:
1. `ToolExecutionEntity` — sets toolOutput, status, durationMs
2. `TraceNodeEntity` — sets completedOn, durationMs, status

### External MCP Servers

External MCP servers (registered via `McpServerEntity`) are **not instrumented** by
this issue. They run their own code and Axiom has no control over their internals.
Tool calls to external MCP servers will only be visible in the execution log text,
not as trace nodes.

Future enhancement: if external MCP servers support OpenTelemetry or similar, their
spans could be correlated via the trace ID.

## Key Files to Modify/Create
- `templates/axiom-mcp-server/trace-helper.js` — **New** shared trace helper
- `templates/axiom-mcp-server/sdk-server.js` — Wrap tool handlers
- `templates/axiom-mcp-server/tools-server.js` — Wrap tool handlers
- `app/src/main/java/io/apitomy/axiom/app/rest/TraceResourceImpl.java` — Implement the
  callback handler methods (the OpenAPI contract and generated interface for these
  endpoints are created in #49)

**Note:** The OpenAPI schema for the callback endpoints is owned by #49. This issue
implements the Java handler logic within the same `TraceResourceImpl` class that #49
creates — NOT a separate `TraceCallbackResource`.

### Non-Fatal Tracing
Tool call tracing is **non-fatal**: if the trace callback fails (network error,
Axiom API down), the tool execution proceeds normally. Trace data is best-effort.
This is critical — tracing must never break tool execution.

### Output Truncation
Tool outputs can be very large (e.g., file contents, API responses). Truncate
`toolOutput` to a reasonable limit (e.g., 10,000 characters) to avoid bloating
the `tool_execution` table.

### No Authentication
The callback endpoint requires no authentication. It runs on localhost and is only
called by MCP servers that Axiom itself spawned. The trace ID acts as a weak
correlation token.

### Concurrency
Multiple tool calls can happen concurrently during a single task/report execution.
Each creates its own trace node independently, all sharing the same parent node ID.
The callback endpoint must handle concurrent requests safely.

## Acceptance Criteria
- [ ] SDK server wraps all tool calls with trace callbacks
- [ ] Tools server wraps all script tool calls with trace callbacks
- [ ] Shared trace-helper.js avoids code duplication
- [ ] Callback REST endpoint creates ToolExecutionEntity + TraceNodeEntity atomically
- [ ] Completion endpoint updates both entities
- [ ] Tool execution works correctly when tracing env vars are absent (graceful no-op)
- [ ] Tool execution works correctly when trace callback fails (non-fatal)
- [ ] Tool outputs are truncated to prevent database bloat
- [ ] Concurrent tool calls create independent trace nodes under same parent
- [ ] Integration test: execute a task with tool calls and verify trace tree

EricWittmann:
## Summary
- Create shared `trace-helper.js` module with `startToolTrace`/`completeToolTrace` functions for non-fatal trace callbacks with 10K char output truncation
- Wrap `sdk-server.js` and `tools-server.js` `CallToolRequestSchema` handlers to emit trace nodes before/after every tool execution
- Add `trace-helper.js` to `McpConfigGenerator.TEMPLATE_FILES` so it gets copied to the MCP server install directory
- Add `@Transactional` to `TraceResourceImpl.createToolCall()` and `completeToolCall()` to ensure Panache writes are flushed

## Related Issue
Fixes #48

## Test Plan
- [ ] `./build.sh` passes (260 tests, 0 failures) including MCP server integration tests
- [ ] Tool execution works correctly when tracing env vars are absent (graceful no-op)
- [ ] Tool execution works correctly when trace callback fails (non-fatal)
- [ ] Manual: execute a task with MCP tools, verify `trace_node` and `tool_execution` rows created

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/McpConfigGenerator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/McpConfigGenerator.java
@@ -51,7 +51,7 @@ public class McpConfigGenerator {
 
     private static final String MCP_SERVER_DIR_NAME = "mcp-server";
     private static final String[] TEMPLATE_FILES = {
-            "package.json", "sdk-server.js", "tools-server.js" };
+            "package.json", "sdk-server.js", "tools-server.js", "trace-helper.js" };
 
     @ConfigProperty(name = "quarkus.http.port", defaultValue = "9090")
     int httpPort;
--- a/app/src/main/java/io/apitomy/axiom/app/rest/TraceResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/TraceResourceImpl.java
@@ -25,6 +25,7 @@ import io.quarkus.panache.common.Sort;
 import io.smallrye.common.annotation.RunOnVirtualThread;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
 import jakarta.ws.rs.WebApplicationException;
 
 import java.math.BigInteger;
@@ -136,6 +137,7 @@ public class TraceResourceImpl implements TracesResource {
     }
 
     @Override
+    @Transactional
     public ToolCallCreated createToolCall(ToolCallRequest data) {
         UUID traceUuid = data.getTraceId();
         TraceEntity trace = TraceEntity.findById(traceUuid);
@@ -168,6 +170,7 @@ public class TraceResourceImpl implements TracesResource {
     }
 
     @Override
+    @Transactional
     public void completeToolCall(BigInteger nodeId, ToolCallCompletion data) {
         TraceNodeEntity node = TraceNodeEntity.findById(nodeId.longValue());
         if (node == null) {
--- a/app/src/main/resources/templates/axiom-mcp-server/sdk-server.js
+++ b/app/src/main/resources/templates/axiom-mcp-server/sdk-server.js
@@ -1,6 +1,7 @@
 const { Server } = require("@modelcontextprotocol/sdk/server/index.js");
 const { StdioServerTransport } = require("@modelcontextprotocol/sdk/server/stdio.js");
 const { ListToolsRequestSchema, CallToolRequestSchema } = require("@modelcontextprotocol/sdk/types.js");
+const { startToolTrace, completeToolTrace } = require(require("path").join(__dirname, "trace-helper.js"));
 
 function log(level, message, data) {
     const entry = {
@@ -367,14 +368,22 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
     }
 
     log("INFO", "SDK tool called", { toolName, args: Object.keys(args) });
+    const toolInput = JSON.stringify(args);
+    const traceNodeId = await startToolTrace(toolName, toolInput);
+
+    let result, status = "success";
+    const startTime = Date.now();
     try {
-        const startTime = Date.now();
-        const result = await tool.handler(args);
+        result = await tool.handler(args);
         const durationMs = Date.now() - startTime;
         log("INFO", "SDK tool completed", { toolName, durationMs });
+        await completeToolTrace(traceNodeId, result, status, durationMs);
         return { content: [{ type: "text", text: result || "OK" }] };
     } catch (error) {
+        status = "failure";
+        const durationMs = Date.now() - startTime;
         log("ERROR", "SDK tool failed", { toolName, error: error.message });
+        await completeToolTrace(traceNodeId, error.message, status, durationMs);
         return { content: [{ type: "text", text: error.message }], isError: true };
     }
 });
--- a/app/src/main/resources/templates/axiom-mcp-server/tools-server.js
+++ b/app/src/main/resources/templates/axiom-mcp-server/tools-server.js
@@ -1,6 +1,7 @@
 const { Server } = require("@modelcontextprotocol/sdk/server/index.js");
 const { StdioServerTransport } = require("@modelcontextprotocol/sdk/server/stdio.js");
 const { ListToolsRequestSchema, CallToolRequestSchema } = require("@modelcontextprotocol/sdk/types.js");
+const { startToolTrace, completeToolTrace } = require(require("path").join(__dirname, "trace-helper.js"));
 const { execSync } = require("child_process");
 const fs = require("fs");
 const os = require("os");
@@ -61,6 +62,8 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
     }
 
     log("INFO", "Script tool called", { toolName, args: Object.keys(args) });
+    const toolInput = JSON.stringify(args);
+    const traceNodeId = await startToolTrace(toolName, toolInput);
 
     try {
         let cmd = tool.scriptTemplate;
@@ -93,10 +96,12 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
         const durationMs = Date.now() - startTime;
 
         log("INFO", "Script tool completed", { toolName, durationMs, outputLength: (result || "").length });
+        await completeToolTrace(traceNodeId, result, "success", durationMs);
         return { content: [{ type: "text", text: result || "Command completed successfully" }] };
     } catch (error) {
         const msg = error.stderr || error.stdout || error.message || "Command failed";
         log("ERROR", "Script tool failed", { toolName, exitCode: error.status, error: msg.substring(0, 500) });
+        await completeToolTrace(traceNodeId, msg, "failure", 0);
         return { content: [{ type: "text", text: msg }], isError: true };
     }
 });
--- a/app/src/main/resources/templates/axiom-mcp-server/trace-helper.js
+++ b/app/src/main/resources/templates/axiom-mcp-server/trace-helper.js
@@ -1,0 +1,54 @@
+const TRACE_ID = process.env.AXIOM_TRACE_ID;
+const PARENT_NODE_ID = process.env.AXIOM_PARENT_NODE_ID;
+const AXIOM_API_URL = process.env.AXIOM_API_URL;
+
+const MAX_OUTPUT_LENGTH = 10000;
+
+const tracingEnabled = !!(TRACE_ID && PARENT_NODE_ID && AXIOM_API_URL);
+
+async function startToolTrace(toolName, toolInput) {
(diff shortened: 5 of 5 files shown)


EricWittmann (commented):
## Code Review: Tool call tracing via MCP server instrumentation

This PR adds tool-call tracing to the MCP servers (`sdk-server.js` and `tools-server.js`) by introducing a shared `trace-helper.js` module that makes non-fatal HTTP callbacks to the Axiom API before and after every tool execution. It also adds `@Transactional` annotations to the Java REST endpoints that persist trace data.

### Findings

#### ⚠️ Warning

**`tools-server.js` — `durationMs` is hardcoded to `0` on failure (line 104)**

In the error path of `tools-server.js`, `durationMs` is passed as `0` to `completeToolTrace` because `startTime` is scoped inside the `try` block and inaccessible in the `catch`:

```javascript
// tools-server.js, line 104
await completeToolTrace(traceNodeId, msg, "failure", 0);
```

By contrast, `sdk-server.js` correctly handles this by declaring `startTime` *before* the try/catch:

```javascript
// sdk-server.js, line 375
const startTime = Date.now();
try {
    result = await tool.handler(args);
    // ...
} catch (error) {
    const durationMs = Date.now() - startTime;  // ← correctly computed
    // ...
}
```

The fix would be to move `const startTime = Date.now();` above the `try` block in `tools-server.js` (e.g., right after line 66), so the catch block can compute actual duration. This matters for accurately recording how long failed tools ran (e.g., tools that hit their 30s timeout).

---

#### 💡 Suggestion

**`trace-helper.js` — `parseInt` without radix (line 18)**

```javascript
parentNodeId: parseInt(PARENT_NODE_ID),
```

Best practice is to always specify the radix: `parseInt(PARENT_NODE_ID, 10)`. While decimal strings parse correctly without it, the explicit radix prevents subtle bugs if the value ever starts with `0` and avoids linter warnings.

---

#### 💡 Suggestion

**`trace-helper.js` — `completeToolTrace` doesn't check response status (lines 39–48)**

`startToolTrace` checks `resp.ok` before parsing the response, but `completeToolTrace` fire-and-forgets without checking the HTTP status:

```javascript
// startToolTrace checks:
if (resp.ok) {
    const data = await resp.json();
    return data.nodeId;
}

// completeToolTrace doesn't:
await fetch(`${AXIOM_API_URL}/traces/tool-calls/${nodeId}`, { ... });
// no resp.ok check, no logging on failure
```

While the non-fatal design is correct, logging a warning on non-2xx responses would help with debugging when trace completion silently fails. Something like:

```javascript
const resp = await fetch(...);
if (!resp.ok) {
    console.error("Trace complete returned", resp.status);
}
```

---

#### 🔍 Nitpick

**`trace-helper.js` — `tracingEnabled` is exported but not imported anywhere**

The module exports `tracingEnabled` alongside the two functions, but neither `sdk-server.js` nor `tools-server.js` imports it. This isn't harmful, but it's dead API surface. Consider either removing it from exports or documenting it as intentionally available for future use.

---

### Overall Assessment

**This PR looks good.** The design is solid — the trace helper cleanly separates tracing concerns, handles the disabled/error cases gracefully (non-fatal, no-op when env vars are absent), and the `@Transactional` additions on the Java side are necessary and correct for Panache entity persistence. The truncation of tool output at 10K chars is a good safeguard.

The main actionable item is the `durationMs: 0` bug in `tools-server.js` error path, which will cause failed tool executions to always report zero duration in traces. The other items are minor improvements.

PR #61, opened 2026-06-25
Add real-time SSE updates to Trace Detail page

EricWittmann:
## Summary

- Fire `trace-updated` SSE events from `TraceService` and `TraceResourceImpl` whenever a trace or
  trace node is created, updated, or completed
- `TraceDetailPage` subscribes to these events and auto-refreshes the directed graph in real-time
- Enable word wrap on tool execution Input/Output code editors in the trace node detail modal
- Increase MCP tool output capture limit from 10K to 100K characters to avoid truncation

## Test plan

- [ ] Start app with `dev.sh`, trigger a report, navigate to its execution trace — graph should
      update in real-time as nodes appear and complete
- [ ] Open a tool execution node detail modal — Output tab content should word-wrap instead of
      being clipped
- [ ] Verify `[SSE] Event received: trace-updated` messages appear in browser console

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/rest/TraceResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/TraceResourceImpl.java
@@ -20,10 +20,12 @@ import io.apitomy.axiom.core.entities.TaskEntity;
 import io.apitomy.axiom.core.entities.ToolExecutionEntity;
 import io.apitomy.axiom.core.entities.TraceEntity;
 import io.apitomy.axiom.core.entities.TraceNodeEntity;
+import io.apitomy.axiom.core.events.SseEvent;
 import io.quarkus.panache.common.Page;
 import io.quarkus.panache.common.Sort;
 import io.smallrye.common.annotation.RunOnVirtualThread;
 import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Event;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import jakarta.ws.rs.WebApplicationException;
@@ -45,6 +47,9 @@ public class TraceResourceImpl implements TracesResource {
 
     @Inject
     ObjectMapper objectMapper;
+
+    @Inject
+    Event<SseEvent> sseEvents;
 
     @Override
     public TraceSearchResults listTraces(BigInteger page, BigInteger limit,
@@ -164,6 +169,8 @@ public class TraceResourceImpl implements TracesResource {
         node.entityId = toolExec.id;
         node.persist();
 
+        sseEvents.fire(SseEvent.traceUpdated(traceUuid));
+
         ToolCallCreated response = new ToolCallCreated();
         response.setNodeId(node.id);
         return response;
@@ -191,6 +198,8 @@ public class TraceResourceImpl implements TracesResource {
                 toolExec.durationMs = data.getDurationMs();
             }
         }
+
+        sseEvents.fire(SseEvent.traceUpdated(node.traceId));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


EricWittmann (commented):
## Code Review Summary

This PR adds real-time SSE notifications for trace updates, so the Trace Detail page auto-refreshes its directed graph as nodes are created/completed. It also enables word wrap on tool execution code editors and increases the MCP tool output capture limit from 10K to 100K characters.

Overall this is a clean, well-structured change. The SSE subscription pattern on the frontend is solid (proper cleanup via unsubscribe, filtering by traceId). I have one concern worth addressing around transaction timing.

---

### ⚠️ Warning

**SSE events fired inside transaction boundary in `completeNode` and `completeTrace`**
`core/src/main/java/io/apitomy/axiom/core/tracing/TraceService.java`

In `createTrace` and `addNode`, the SSE event is correctly fired **after** the `QuarkusTransaction.requiringNew()` lambda returns (i.e. after the transaction commits):

```java
// createTrace — correct: event fires after commit
TraceContext ctx = QuarkusTransaction.requiringNew().call(() -> { ... });
sseEvents.fire(SseEvent.traceUpdated(ctx.traceId()));
return ctx;
```

However, in both `completeNode` overloads and `completeTrace`, the SSE event is fired **inside** the transaction lambda — before the transaction commits:

```java
// completeNode — event fires before commit
QuarkusTransaction.requiringNew().run(() -> {
    // ... update node ...
    sseEvents.fire(SseEvent.traceUpdated(node.traceId)); // ← inside lambda
});
```

Since CDI events are synchronous, the SSE broadcast reaches clients before the transaction commits. A fast client receiving the event and immediately re-fetching the trace could see stale data (the old node status). This is the same race condition that `createTrace`/`addNode` avoid by firing after the lambda.

**Suggested fix:** Restructure these methods to capture the `traceId` from inside the lambda and fire the event after:

```java
public void completeNode(Long nodeId, String status) {
    UUID traceId = QuarkusTransaction.requiringNew().call(() -> {
        TraceNodeEntity node = TraceNodeEntity.findById(nodeId);
        if (node == null) {
            LOG.warnf("Trace node %d not found for completion", nodeId);
            return null;
        }
        Instant now = Instant.now();
        node.completedOn = now;
        node.durationMs = Duration.between(node.startedOn, now).toMillis();
        node.status = status;
        return node.traceId;
    });
    if (traceId != null) {
        sseEvents.fire(SseEvent.traceUpdated(traceId));
    }
}
```

Same pattern would apply to the other `completeNode` overload and `completeTrace`.

The same issue technically applies to `createToolCall` and `completeToolCall` in `TraceResourceImpl.java` (the `@Transactional` interceptor commits after the method returns, so the SSE event fires before commit), but the window is much smaller there and less likely to cause visible issues.

---

### 💡 Suggestion

**Consider debouncing SSE-triggered refreshes**
`ui/src/pages/TraceDetailPage.tsx`

During an active trace execution, many nodes may be created and completed in rapid succession. Each SSE event increments `refreshKey`, triggering a full data re-fetch + ELK layout recalculation. For a trace with, say, 20 tool calls running near-simultaneously, this could produce 40+ rapid re-renders.

Consider debouncing or throttling the refresh — e.g., coalescing updates that arrive within a short window (200–500ms):

```typescript
useEffect(() => {
    let timeout: ReturnType<typeof setTimeout>;
    const unsubscribe = sseClient.subscribe((event: AxiomSseEvent) => {
        if (event.type === "trace-updated") {
            const data = event.data as { traceId?: string };
            if (data.traceId === traceId) {
                clearTimeout(timeout);
                timeout = setTimeout(() => setRefreshKey((k) => k + 1), 300);
            }
        }
    });
    return () => { clearTimeout(timeout); unsubscribe(); };
}, [traceId]);
```

This isn't critical for correctness but would improve performance and reduce flicker during heavy trace activity.

---

### 🔹 Nitpick

**Fully-qualified `java.util.UUID` in method signature**
`core/src/main/java/io/apitomy/axiom/core/events/SseEvent.java`

The `traceUpdated` method uses the fully-qualified `java.util.UUID` type instead of adding an import:

```java
public static SseEvent traceUpdated(java.util.UUID traceId) {
```

This works but is inconsistent with typical Java style. Adding `import java.util.UUID;` at the top and using `UUID traceId` in the signature would be cleaner.

---

### Overall Assessment

**Looks good with minor changes.** The core approach is sound — the SSE plumbing, client-side subscription, and filtering are all well done. The main item to address is the inconsistency in when SSE events are fired relative to transaction commit in `completeNode`/`completeTrace`, which could cause intermittent stale reads on the client. The debouncing suggestion is a nice-to-have for polish.

PR #62, opened 2026-06-25
Split pipeline transaction to avoid holding DB connection during AI call (#44)

### LINKED ISSUE #44: Long-running database transaction held open during AI engine call in processNextEvent()
## Problem

`PipelineOrchestrator.processNextEvent()` is annotated with `@Transactional` (added in
#31 / PR #43 to fix a CDI self-invocation bug). This means the entire pipeline — including
the AI engine call (`aiEngine.promptWithSchema().join()`) in `ManagerService.evaluate()` —
runs inside a single JTA transaction.

The AI call can block for up to `axiom.manager.timeout-seconds` (default **120 seconds**),
holding a database connection from the pool for the full duration. While
`ConcurrentExecution.SKIP` prevents concurrent scheduler invocations, a 120-second
connection hold is still significant and could lead to pool contention if other parts of the
application need connections during that window.

## Affected Files

- `app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java` — `processNextEvent()`
- `manager/src/main/java/io/apitomy/axiom/manager/ManagerService.java` — `evaluate()`

## Proposed Fix

Split the pipeline into three transactional phases so the AI call runs outside any
transaction:

1. **Phase 1 — Dequeue** (short transaction): Find the next pending event, mark it as
   `"processing"`, and commit. This prevents double-processing.
2. **Phase 2 — AI Call** (no transaction): Invoke `managerService.evaluate()` with the
   event. The AI engine call blocks here, but no database connection is held.
3. **Phase 3 — Persist results** (short transaction): Process the Manager's decisions
   (create tasks, log activity, mark queue entry as `"completed"` or `"failed"`).

This requires extracting `findNextPendingEvent()` into its own transactional boundary
(separate bean or programmatic transaction) and restructuring `processEvent()` so that the
persistence operations after the AI call also run in their own transaction.

`ManagerService.recordAiUsage()` and `logManagerActivity()` already use
`QuarkusTransaction.requiringNew()` (independent transactions), so they would continue to
work correctly in either approach.

## Impact

- **Current risk**: Low — H2 file-based database with a single scheduled consumer. Pool
  contention is unlikely in the current deployment.
- **Future risk**: Higher if Axiom moves to a shared database (e.g. PostgreSQL) with
  multiple concurrent consumers or a larger connection pool under load.

EricWittmann:
## Summary

- **PipelineOrchestrator**: Removed the single `@Transactional` wrapping the entire pipeline
  and replaced it with three short programmatic transactions using
  `QuarkusTransaction.requiringNew()`: (1) dequeue, (2) AI call with no transaction held,
  (3) persist results. Extracted `persistResults()` and `persistFailure()` helper methods.
- **ManagerService**: Wrapped the context-loading JPA reads at the start of `evaluate()`
  in their own short transaction via a new `EvalContext` record, so the database connection
  is released before the AI engine call blocks.
- **PipelineOrchestratorTest**: Adapted to the new transaction model — test data is now
  committed via `QuarkusTransaction.requiringNew()` (matching the `TraceServiceTest` pattern)
  with `@AfterEach` cleanup instead of implicit `@Transactional` rollback.

## Related Issue

Fixes #44

## Test Plan

- [ ] Run `PipelineOrchestratorTest` — all 11 existing tests pass with the refactored transaction
      boundaries
- [ ] Run the full test suite to check for regressions
- [ ] Verify structurally: `processNextEvent()` no longer has `@Transactional`, and the AI call in
      `evaluate()` runs between committed short transactions with no connection held

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -191,6 +191,12 @@ public class PipelineOrchestrator {
     private void persistFailure(Long queueEntryId, Long eventId,
             String errorMessage, TraceContext traceCtx) {
         QuarkusTransaction.requiringNew().run(() -> {
+            if (traceCtx != null) {
+                EventEntity event = EventEntity.findById(eventId);
+                if (event != null) {
+                    event.traceId = traceCtx.traceId();
+                }
+            }
             logActivity(null, null, eventId, "pipeline-error",
                     "Pipeline error: " + errorMessage);
             completeTraceFailed(traceCtx);
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.apitomy.axiom.core.tracing.TraceContext;
 import io.apitomy.axiom.core.tracing.TraceService;
-import io.apitomy.axiom.core.entities.ActionTypeEntity;
 import io.apitomy.axiom.core.entities.ActivityLogEntity;
 import io.apitomy.axiom.core.entities.EventEntity;
 import io.apitomy.axiom.core.entities.EventQueueEntity;
@@ -20,7 +19,7 @@ import io.quarkus.scheduler.Scheduled;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.event.Event;
 import jakarta.inject.Inject;
-import jakarta.transaction.Transactional;
+import io.quarkus.narayana.jta.QuarkusTransaction;
 import org.jboss.logging.Logger;
 
 import java.time.Instant;
@@ -62,8 +61,15 @@ public class PipelineOrchestrator {
     /**
      * Polls the event queue every 5 seconds and processes one pending event at a time.
      * Events are processed sequentially to avoid race conditions in the Manager.
+     *
+     * <p>The pipeline is split into three transactional phases so that the AI engine
+     * call in {@code ManagerService.evaluate()} does not hold a database connection:</p>
+     * <ol>
+     *   <li><b>Dequeue</b> — short transaction to mark the next pending event as "processing"</li>
+     *   <li><b>AI call</b> — no transaction held while the Manager evaluates the event</li>
+     *   <li><b>Persist results</b> — short transaction to process decisions and mark completion</li>
+     * </ol>
      */
-    @Transactional
     @Scheduled(every = "${axiom.pipeline.poll-interval:5s}",
                concurrentExecution = Scheduled.ConcurrentExecution.SKIP)
     void processNextEvent() {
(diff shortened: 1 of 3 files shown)


EricWittmann (commented):
## Review Summary

This PR splits the single long `@Transactional` annotation on the pipeline event processing into three short programmatic transactions using `QuarkusTransaction.requiringNew()`, ensuring the AI engine call in `ManagerService.evaluate()` does not hold a database connection. The same pattern is applied inside `ManagerService` itself, where context loading is wrapped in a short transaction before the AI call. Tests are adapted to match (committed setup data with `@AfterEach` cleanup).

Well-structured change — the three-phase approach (dequeue → AI call → persist) is clean and the detached entity usage is safe given that all entity classes involved use simple scalar fields (no lazy-loaded relationships that would cause `LazyInitializationException`).

---

## Findings

### ⚠️ Warning — `event.traceId` not set on failure path

**File:** `PipelineOrchestrator.java` — `processEvent()` / `persistFailure()`

In the old code, `event.traceId = traceCtx.traceId()` was set immediately after trace creation (before the try/catch), so it was committed regardless of whether the AI call succeeded or failed. In the new code, `event.traceId` is only set inside `persistResults()` (the success path). On the failure path, `persistFailure()` never updates `event.traceId`, so failed events will no longer link back to their trace.

The reverse link (Trace → Event via `trace.eventId`) still works, but the Event → Trace navigation is lost for failures. Consider adding the `traceId` assignment to `persistFailure`:

```java
private void persistFailure(Long queueEntryId, Long eventId,
        String errorMessage, TraceContext traceCtx) {
    QuarkusTransaction.requiringNew().run(() -> {
        if (traceCtx != null) {
            EventEntity event = EventEntity.findById(eventId);
            if (event != null) {
                event.traceId = traceCtx.traceId();
            }
        }
        logActivity(null, null, eventId, "pipeline-error",
                "Pipeline error: " + errorMessage);
        completeTraceFailed(traceCtx);
        markQueueEntry(queueEntryId, "failed");
    });
}
```

### 💡 Suggestion — `TraceService` calls nested inside `persistResults` transaction

**File:** `PipelineOrchestrator.java` — `persistResults()`

Inside the `persistResults` transaction, calls like `popEvalNode(traceCtx)` and `completeTraceIfSync(traceCtx, ...)` eventually call `TraceService` methods which each open their own `QuarkusTransaction.requiringNew()`. This suspends the outer transaction for each inner one and then resumes it. While Narayana handles nested `requiringNew()` correctly, it does create a lot of transaction churn in the persist phase.

This is not a bug and is consistent with how `TraceService` is designed (always self-transactional), but worth being aware of from a performance perspective — each `TraceService` call suspends the outer `persistResults` transaction, commits its own, then resumes. For the typical single-decision case this is fine, but with many decisions it multiplies.

### ✅ Nitpick — `processEvent` Phase 1 could be folded into `findNextPendingEvent`

**File:** `PipelineOrchestrator.java` — `processEvent()`

Phase 1 in `processEvent` opens a short transaction just to load the `EventEntity` by ID. Since `findNextPendingEvent` already opens a short transaction that has access to `queueEntry.eventId`, the event could potentially be loaded there to eliminate one transaction round-trip. However, this is a minor optimization and the current structure is arguably clearer with the separation, so no action needed.

---

## Overall Assessment

**Looks good to merge** (with one minor fix recommended). The core goal — releasing the DB connection during AI calls — is achieved cleanly. The entity usage after detachment is safe (all scalar fields, `ProjectEntity.labels` is `EAGER`), the test migration follows the established `TraceServiceTest` pattern, and the error handling structure is preserved. The only behavioral change worth addressing is the missing `event.traceId` assignment on the failure path (warning above).

PR #66, opened 2026-06-29
Fix race condition in script execution project lookup

EricWittmann:
## Summary

- Fix a race condition where `ScriptExecutionService` re-reads the project from the database
  inside an async `CompletableFuture`, but the enclosing transaction (in `PipelineOrchestrator`)
  hasn't committed yet — causing `{{repository}}` and other project template variables to resolve
  to empty strings
- Add an overloaded `executeScript(task, project)` that accepts the in-memory `ProjectEntity`
  directly, falling back to a DB lookup when `null`
- `PipelineOrchestrator.handleScriptAction` now passes its already-loaded project to avoid the race

## Test plan

- [ ] Trigger an `issue-created` event for a new issue (no existing project) and verify the
      `clone-github-repo` script action resolves `{{repository}}` correctly
- [ ] Verify existing `TaskExecutionService` path (no project argument) still works for
      re-executed script tasks

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -409,7 +409,7 @@ public class PipelineOrchestrator {
         sseEvents.fire(SseEvent.projectUpdated(project.id));
         sseEvents.fire(SseEvent.threadEntry(project.id));
 
-        scriptExecutionService.executeScript(task);
+        scriptExecutionService.executeScript(task, project);
 
         return decisionLog.id;
     }
--- a/app/src/main/java/io/apitomy/axiom/app/ScriptExecutionService.java
+++ b/app/src/main/java/io/apitomy/axiom/app/ScriptExecutionService.java
@@ -60,12 +60,26 @@ public class ScriptExecutionService {
      * @param task the task whose action type defines the script to run
      */
     public void executeScript(TaskEntity task) {
+        executeScript(task, null);
+    }
+
+    /**
+     * Executes the script for a task asynchronously, using the supplied project
+     * for template variable resolution. If {@code project} is {@code null}, the
+     * project is loaded from the database — callers that already hold a managed
+     * entity (e.g. inside an uncommitted transaction) should pass it directly to
+     * avoid a race between the async thread and the transaction commit.
+     *
+     * @param task    the task whose action type defines the script to run
+     * @param project the project to use for placeholder substitution, or {@code null}
+     */
+    public void executeScript(TaskEntity task, ProjectEntity project) {
         markTaskInProgress(task.id);
 
         CompletableFuture.runAsync(() -> {
             Arc.container().requestContext().activate();
             try {
-                RunResult result = runScript(task);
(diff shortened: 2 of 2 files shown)


EricWittmann (commented):
## Code Review — Fix race condition in script execution project lookup

**Summary:** This PR fixes a race condition where `ScriptExecutionService.runScript()` re-reads the project from the database inside a `CompletableFuture.runAsync()` block, but the enclosing `QuarkusTransaction.requiringNew()` in `PipelineOrchestrator.persistResults()` hasn't committed yet — causing `{{repository}}` and other project-scoped template variables to resolve to empty strings (or worse, an NPE if the project lookup returns `null`). The fix adds an overloaded `executeScript(task, project)` method that accepts the in-memory entity directly, with a backward-compatible no-arg overload for callers like `TaskExecutionService` where the project is already committed.

---

### Findings

No critical, warning-level, or significant issues found. This is a clean, well-scoped fix.

<details>
<summary><b>🟢 Nitpick</b> — Pre-existing: <code>markTaskInProgress</code> self-call bypasses <code>@Transactional</code></summary>

**File:** `ScriptExecutionService.java:77`

`executeScript()` calls `markTaskInProgress(task.id)`, which is a self-invocation within the same CDI bean. CDI interceptors (including `@Transactional`) are not applied on self-calls. Today this works because `executeScript` is always called from within an active transaction context (`PipelineOrchestrator.persistResults` → `QuarkusTransaction.requiringNew()`), so the method participates in the outer transaction. However, if `executeScript` were ever called from a non-transactional context, `markTaskInProgress` would silently run without a transaction.

This is pre-existing and not introduced by this PR — just noting it for awareness. The same pattern exists in `completeTask` and `failTask` calls from within the async `CompletableFuture`, but those explicitly use `Arc.container().requestContext()` and Quarkus provides implicit transaction management for Panache operations in that context.

</details>

---

### What looks good

- **Correct root cause identification.** The race between `CompletableFuture.runAsync()` and the uncommitted `QuarkusTransaction.requiringNew()` is a real and subtle bug. Passing the in-memory entity is the right fix.

- **Thread-safety of the approach.** The `ProjectEntity` fields used in `substitutePlaceholders` (`issueRef`, `repository`, `name`) are all simple `String` columns — no lazy-loaded associations. The `@ElementCollection(fetch = EAGER)` on `labels` means the entity is fully initialized. By the time the async thread reads these fields, the calling thread has finished constructing the object. No mutation races.

- **Backward compatibility preserved.** The single-arg `executeScript(TaskEntity)` overload delegates to the two-arg version with `null`, maintaining the existing `TaskExecutionService.executeTask()` call path (line 116) where the project is already committed to the DB.

- **Null-safe fallback.** `runScript` falls back to `ProjectEntity.findById(task.projectId)` when `project` is `null`, so the existing code path is unchanged.

- **Javadoc quality.** The new overload's documentation clearly explains when and why callers should pass the project directly vs. relying on the DB lookup. The commit message body is equally thorough.

---

### Overall Assessment

✅ **Looks good to merge.** This is a focused, minimal fix for a real race condition. The approach is correct, the backward-compatible overload is clean, and no new risks are introduced.

PR #88, opened 2026-07-16
Add browser notifications for assistant permission requests

EricWittmann:
## Summary

- Fire desktop notifications via the browser Notification API when an assistant session needs
  user action (tool permission, plan approval, or a question) and the Axiom tab is not focused
- Request notification permission once when the chat panel mounts
- Clicking a notification brings the Axiom tab to the foreground
- Uses `tag` to deduplicate notifications for the same permission request

## Test plan

- [ ] Open an assistant session, switch to a different tab, and trigger a tool that requires
      permission — verify a desktop notification appears
- [ ] Click the notification and verify the Axiom tab is focused
- [ ] Verify no notification fires when the Axiom tab is already in the foreground
- [ ] Verify the browser permission prompt appears on first use

### DIFF
--- a/ui/src/components/assistant/AssistantChatPanel.tsx
+++ b/ui/src/components/assistant/AssistantChatPanel.tsx
@@ -33,6 +33,7 @@ export function AssistantChatPanel({ sessionId, onItemsChanged, onModeChange, on
     const reconnectTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
     const reconnectDelayRef = useRef(1000);
     const sessionEndedRef = useRef(false);
+    const activeNotificationsRef = useRef<Map<string, Notification>>(new Map());
 
     // Keep callback refs so the EventSource effect doesn't re-run when
     // parent-supplied callbacks change identity (e.g. inline arrow functions).
@@ -196,7 +197,7 @@ export function AssistantChatPanel({ sessionId, onItemsChanged, onModeChange, on
                     if (document.hidden
                             && "Notification" in window
                             && Notification.permission === "granted") {
-                        const toolName = data.toolName as string;
+                        const toolName = data.toolName;
                         let title = "Action required";
                         let body = `${toolName} needs your approval`;
                         if (toolName === "ExitPlanMode") {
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review — Add browser notifications for assistant permission requests

This PR adds browser desktop notifications (via the Notification API) when a permission request arrives while the Axiom tab is not focused. It requests permission on mount, fires contextual notifications for different tool types, and uses `tag`-based deduplication. Clean, well-scoped change.

---

### ⚠️ Warning

**Notification not dismissed when the permission is resolved**
`AssistantChatPanel.tsx` — `permission_request` handler vs. `permission_resolved` handler

When a `permission_resolved` event arrives (e.g. the user switched back to the tab and approved/denied, or it was resolved from another session), the notification that was fired remains visible in the OS notification center. Consider storing a reference to the notification and closing it when the matching `permission_resolved` event is received:

```ts
// e.g. in a ref
const activeNotificationsRef = useRef<Map<string, Notification>>(new Map());

// In permission_request handler:
activeNotificationsRef.current.set(data.requestId, notification);

// In permission_resolved handler:
activeNotificationsRef.current.get(data.permissionId)?.close();
activeNotificationsRef.current.delete(data.permissionId);
```

This prevents stale notifications from lingering and avoids confusing the user if they click a notification for an already-resolved permission.

---

### 💡 Suggestion

**`Notification.requestPermission()` on mount may be silently blocked**
`AssistantChatPanel.tsx`, lines 52–56

Modern browsers (Chrome 110+, Firefox 72+) increasingly require a **user gesture** (click, keypress) to trigger the notification permission prompt. Calling `requestPermission()` inside a `useEffect` on mount may be silently ignored by the browser — the prompt simply won't appear, and the permission remains `"default"`.

This isn't a bug (it degrades gracefully — notifications just won't fire), but it means the user may never see the permission prompt. Consider deferring the permission request to the first time a permission_request event is received, or triggering it from a user-initiated action (e.g. a "Enable notifications" button or on first message send).

---

### 🔍 Nitpick

**Unnecessary `as string` cast**
`AssistantChatPanel.tsx`, line 199

```ts
const toolName = data.toolName as string;
```

Since `data` is already `any` (from `JSON.parse`), the `as string` cast doesn't add type safety — `data.toolName` can be assigned to a `string` variable without it. It's harmless, but the cast could give a false sense of safety. Either remove it or add a runtime guard (`typeof data.toolName === "string"`) if you want actual safety.

---

### Overall Assessment

**Looks good — minor changes recommended.** The core logic is correct: the `document.hidden` check, the `tag`-based deduplication, and the `onclick` handler are all well done. The notification content is nicely contextualized for different tool types. The two substantive suggestions above (dismissing notifications on resolution and the user-gesture concern) are worth addressing but neither is a blocker. Clean PR overall.

PR #89, opened 2026-07-16
Show all templates in session filter dropdown

### LINKED ISSUE #82: Template filter dropdown only shows templates with active sessions
## Description

On the AI Assistant session list page, the template filter dropdown only shows templates for which
active sessions currently exist. For example, if only General Assistant sessions are running, the
dropdown shows "All templates" and "General Assistant" — "Axiom Configuration Assistant" is absent.

## Problem

This can be confusing — a user might think templates are missing or misconfigured when they're
simply not represented by any active session.

## Suggested Improvement

Options (pick one):
1. **Show all templates always** — Show every available template in the dropdown, even if no
   sessions use that template. Templates with no matching sessions would simply show an empty
   filtered list.
2. **Show all with counts** — Show all templates with a count indicator, e.g.,
   "General Assistant (2)" and "Axiom Configuration Assistant (0)".
3. **Grey out empty templates** — Show all templates but visually dim those with no active sessions.

Option 1 is simplest and probably sufficient.

EricWittmann:
## Summary
- Fixes the template filter dropdown on the AI Assistant session list page to show all available
  templates, not just those with active sessions
- Templates are now sorted alphabetically by name for consistent ordering
- Closes #82

## Details
The dropdown derived its options from `sessions` (only templates with active sessions). Changed it
to use `templateMap` (already populated from `fetchAssistantTemplates()` with all templates). No API
changes needed — the data was already being fetched but not used for the dropdown.

## Test plan
- [ ] Navigate to the AI Assistant page with sessions using only some templates
- [ ] Confirm the filter dropdown lists all available templates, including those with no sessions
- [ ] Select a template with no sessions — verify it shows an empty filtered list
- [ ] Select a template with sessions — verify filtering still works correctly
- [ ] Verify "All templates" option still shows all sessions

### DIFF
--- a/ui/src/pages/AssistantPage.tsx
+++ b/ui/src/pages/AssistantPage.tsx
@@ -157,7 +157,8 @@ export function AssistantPage() {
         return true;
     });
 
-    const uniqueTemplateIds = [...new Set(sessions.map((s) => s.templateId))];
+    const uniqueTemplateIds = Object.keys(templateMap)
+        .sort((a, b) => (templateMap[a] || a).localeCompare(templateMap[b] || b));
 
     const [deleteTarget, setDeleteTarget] = useState<string | null>(null);
 


EricWittmann (commented):
## Code Review Summary

This PR fixes the template filter dropdown on the AI Assistant session list page to show **all** available templates (sourced from `templateMap`) instead of only those with active sessions (previously derived from `sessions`). It also adds alphabetical sorting by template name.

## Analysis

**Correctness:** ✅ The fix is clean and correct. `templateMap` is already populated from `fetchAssistantTemplates()` in the `load` callback, so the data source switch requires no API changes. `Object.keys()` naturally produces unique IDs, and the fallback `templateMap[x] || x` in the sort comparator is defensive even though values will always exist (since the keys come from the same map).

**Edge cases:** ✅ 
- If `templateMap` is empty (initial state before load completes), `Object.keys({})` returns `[]` — no issues.
- The toolbar is only rendered when `!loading && sessions.length > 0`, so by the time the dropdown is visible, `templateMap` has been populated.
- Selecting a template with no sessions will correctly show an empty `filteredSessions` list, since the filter logic at line 150–158 is unchanged and works independently.

**Performance:** ✅ `Object.keys()` + `.sort()` on a small template list is negligible. This runs on every render (no `useMemo`), but that's consistent with the existing pattern for `filteredSessions` on line 150 and the prior `uniqueTemplateIds` computation — no concern here.

**Code quality:** ✅ The change is minimal and follows existing patterns in the file.

## Findings

No issues found. This is a well-scoped, minimal fix that correctly addresses the reported bug.

## Verdict

✅ **Looks good to merge.** Clean one-line data source change with proper sorting — no concerns.

PR #93, opened 2026-07-21
Replace volatile fields with atomic types in cost accumulation

### LINKED ISSUE #75: AI Assistant: Cost accumulation uses non-atomic operations on volatile fields
## Description

The `accumulateCost` method in `AssistantSession` uses `+=` on volatile `double` and `long` fields.
Since only one thread (the stdout reader) calls this method, the code is safe in practice. However,
the single-writer contract is implicit — using `AtomicLong`/`DoubleAdder` would make the intent
explicit and survive future refactors that might introduce additional writers.

## Location

`app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java` lines 487-493:

```java
private void accumulateCost(SseEvent event) {
    totalCostUsd += event.data().path("costUsd").asDouble(0);
    totalInputTokens += event.data().path("inputTokens").asLong(0);
    totalOutputTokens += event.data().path("outputTokens").asLong(0);
    totalDurationMs += event.data().path("durationMs").asLong(0);
    turnCount++;
}
```

## Suggested Fix

Replace the volatile fields with `AtomicLong` (for token counts, duration, turn count) and
`DoubleAdder` (for cost). This makes thread safety guarantees explicit without changing behavior.

EricWittmann:
## Summary
- Replaced non-atomic compound operations (`+=`, `++`) on `volatile` primitive fields in
  `AssistantSession.accumulateCost()` with proper atomic types (`DoubleAdder`, `AtomicLong`,
  `AtomicInteger`)
- Getter return types unchanged — no caller modifications needed

## Test plan
- [ ] Compile the project (`mvn install` or `build.sh`)
- [ ] Run an assistant session and verify cost/token/turn metrics accumulate correctly
- [ ] Verify session info endpoint returns expected values

Closes #75

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java
+++ b/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java
@@ -17,7 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.DoubleAdder;
 import java.util.function.Consumer;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Review Summary

This PR replaces `volatile` primitive fields (`double`, `long`, `int`) used for cost/token/turn accumulation in `AssistantSession` with proper atomic types (`DoubleAdder`, `AtomicLong`, `AtomicInteger`). This addresses the fact that compound operations like `+=` and `++` on `volatile` fields are **not** atomic — they involve a read, modify, and write that can race under concurrent access.

## Findings

### ✅ No significant issues found

The change is clean, focused, and correct. Some specific observations:

1. **Correct type choices:** `DoubleAdder` is the ideal choice for `totalCostUsd` — it's designed for high-contention addition with better throughput than `AtomicDouble` (which doesn't exist in `java.util.concurrent` anyway). `AtomicLong` and `AtomicInteger` are appropriate for the other counters.

2. **`final` fields:** All five fields are now `final` references to mutable atomic objects — this is the correct pattern (the reference is safely published via `final`; the atomic type handles the value's thread-safety).

3. **Getter return types preserved:** The getters still return `double`, `long`, and `int` primitives, so callers are unaffected.

4. **`accumulateCost` method:** The atomic operations (`add`, `addAndGet`, `incrementAndGet`) are individually atomic. While the five updates are not collectively atomic with each other, that's fine here — the previous `volatile` implementation had the same non-atomicity across fields, and callers reading these values only need eventual consistency per-field.

### 💡 Suggestion (minor, non-blocking)

- **`addAndGet` vs `getAndAdd`:** In `accumulateCost()`, the return values of `addAndGet()` are discarded. You could use `getAndAdd()` instead (which is semantically equivalent when the return value is unused), but this is purely stylistic — there's no performance difference. No change needed.

## Overall Assessment

**Looks good to merge.** This is a straightforward, well-scoped improvement that makes thread-safety guarantees explicit rather than relying on an implicit single-writer contract. The implementation is correct and follows standard Java concurrency idioms.

PR #97, opened 2026-07-21
Allow any template for project assistant sessions

EricWittmann:
## Summary

- Extract the template picker and session name modals from `AssistantPage` into a shared
  `CreateSessionModal` component
- Replace the hardcoded `project-assistant` template on the project detail page with the
  template picker, allowing users to choose any configured template for project sessions
- The backend already injects project context (workspace, system prompt, env vars, MCP tools)
  regardless of which template is used, so no backend changes are needed

## Test plan

- [ ] On the Assistant page, click "New Session" — verify the template picker and name modal
      work as before
- [ ] On a project detail page, click "Assistant" — verify the template picker appears with all
      templates, select one, name the session, and verify it opens in a breakout window
- [ ] Create a project session with the "General Assistant" template — verify project context
      is still injected (system prompt, workspace directory)
- [ ] Cancel the template picker — verify no session is created

### DIFF
--- a/ui/src/components/assistant/CreateSessionModal.tsx
+++ b/ui/src/components/assistant/CreateSessionModal.tsx
@@ -1,0 +1,202 @@
+import { useState, useEffect, useRef } from "react";
+import {
+    Alert,
+    Button,
+    Form,
+    FormGroup,
+    Modal,
+    ModalBody,
+    ModalFooter,
+    ModalHeader,
+    SearchInput,
+    TextInput,
+} from "@patternfly/react-core";
+import {
+    createAssistantSession,
+    fetchAssistantTemplates,
+    type AssistantSessionInfo,
+    type SessionTemplate,
+} from "../../config/api";
+import "../../pages/AssistantPage.css";
+
+const FUN_WORDS = [
+    "rocket", "cactus", "penguin", "waffle", "thunder", "mango", "cosmic",
+    "pickle", "turbo", "noodle", "galaxy", "biscuit", "phantom", "pretzel",
+    "velvet", "zigzag", "bamboo", "coral", "doodle", "falcon", "gopher",
+    "cobalt", "marble", "nimbus", "orchid", "quartz", "walrus", "tundra",
+    "nebula", "pebble", "saffron", "breeze", "mosaic", "lantern", "crimson",
+    "meadow", "jasper", "harbor", "ember", "frost", "summit", "canyon",
+    "copper", "willow", "sparrow", "clover", "rapids", "flint", "comet",
+    "puzzle", "sphinx", "goblin", "dragon", "wizard", "pirate", "ninja",
+    "viking", "yeti", "kraken", "phoenix", "griffin", "titan", "tempest",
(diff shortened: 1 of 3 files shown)


EricWittmann (commented):
## Review Summary

This PR extracts the template picker and session name modals from `AssistantPage` into a reusable `CreateSessionModal` component, then uses it in both `AssistantPage` and `ProjectDetailPage`. The key behavioral change is that the project detail page now shows the full template picker instead of hardcoding the `project-assistant` template, allowing users to choose any configured template for project sessions. Clean refactoring with a well-designed component interface.

---

### 🟡 Suggestions

**1. Template list items lack keyboard accessibility**
`CreateSessionModal.tsx` — lines 138–153

The template items are plain `<div>` elements with only `onClick`. Users navigating by keyboard cannot select a template. Consider adding `role="button"`, `tabIndex={0}`, and an `onKeyDown` handler for Enter/Space:

```tsx
<div
    key={t.templateId}
    className="axiom-assistant-page__template-item"
    role="button"
    tabIndex={0}
    onClick={() => handleTemplateSelect(t)}
    onKeyDown={(e) => {
        if (e.key === "Enter" || e.key === " ") handleTemplateSelect(t);
    }}
>
```

This is consistent with how the session name input already handles Enter-key submission.

**2. Component imports page-level CSS**
`CreateSessionModal.tsx` — line 20

```ts
import "../../pages/AssistantPage.css";
```

The shared component reaches into `pages/` for its styles, and the CSS class names are still prefixed `axiom-assistant-page__`. This works today but creates a coupling — if someone reorganizes or removes the page CSS, the component breaks silently. Consider either:
- Moving the template-picker and create-error CSS rules into a co-located `CreateSessionModal.css` with appropriately renamed classes, or
- Leaving it as-is with a comment noting the shared styles origin

This is a minor concern for a codebase of this size, not blocking.

---

### 🔵 Nitpicks

**3. No "Back" affordance from the name modal**
`CreateSessionModal.tsx`

When a user picks a template and lands on the "Name Your Session" modal, the only options are "Create Session" or "Cancel" (which exits the entire flow). A "Back" button to return to the template picker could be a nice UX touch — but this matches the prior behavior in `AssistantPage`, so it's not a regression.

---

### ✅ What looks good

- **Component interface is clean**: `isOpen`, `onClose`, `onSessionCreated`, plus optional `projectId` and `defaultName` — easy to understand and use from both call sites.
- **State is properly encapsulated**: The modal manages its own template fetching, filtering, and creation flow internally. Parent components only deal with open/close and the created session.
- **Faithful extraction**: The logic moved from `AssistantPage` to `CreateSessionModal` is essentially identical, minimizing regression risk.
- **`ProjectDetailPage` correctly passes `projectId` and `defaultName`**: The `createAssistantSession` API already accepts `projectId` as an optional parameter, so this flows through correctly.
- **Unused imports are cleaned up**: `Alert`, `useRef`, `createAssistantSession`, and `SessionTemplate` are properly removed from `AssistantPage`.

---

### Overall Assessment

**Looks good to merge.** This is a well-executed refactoring. The one substantive suggestion (keyboard accessibility on template items) would be worth addressing but isn't blocking.

PR #98, opened 2026-07-21
Fix SSE reconnection in AssistantChatPanel

### LINKED ISSUE #71: AI Assistant: SSE reconnection is broken
## Description

The `AssistantChatPanel` SSE connection has no reconnection logic. When the `EventSource` encounters
an error (network hiccup, server restart, etc.), the `onerror` handler simply closes the connection
and stops processing — silently killing the entire session stream.

## Location

`ui/src/components/assistant/AssistantChatPanel.tsx` lines 219-222:

```typescript
es.onerror = () => {
    es.close();
    setIsProcessing(false);
};
```

## Impact

A transient network interruption will permanently disconnect the user from an active session with no
indication or recovery. The backend already supports full event history replay on reconnect, so the
infrastructure to fix this is already in place.

## Suggested Fix

Implement exponential backoff reconnection in `onerror`. On reconnect, the SSE endpoint will replay
all buffered events, so the UI will catch up automatically. Consider showing a "Reconnecting..."
indicator in the chat panel during reconnection attempts.

EricWittmann:
## Summary

- Add exponential backoff reconnection (1s initial, 30s max, 10 attempts) to the assistant chat
  SSE connection, which previously silently died on any connection error
- Skip already-processed events during backend replay on reconnect to prevent duplicate messages
- Show an inline "Connection lost. Reconnecting..." warning alert during reconnection attempts
- Display a "Connection lost" system message after max retries are exhausted

Closes #71

## Test plan

- [ ] Open the AI Assistant chat panel and start a conversation
- [ ] Simulate a network interruption (e.g. toggle network off/on, or restart the backend)
- [ ] Verify the "Reconnecting..." warning appears and disappears once reconnected
- [ ] Verify the conversation continues without duplicate messages after reconnect
- [ ] Verify that after 10 failed reconnection attempts, a "Connection lost" system message appears
- [ ] Verify normal operation (no disconnection) is unaffected


EricWittmann (commented):
## Code Review — Fix SSE reconnection in AssistantChatPanel

This PR adds exponential backoff reconnection to the assistant chat SSE connection, with a replay-skip mechanism to deduplicate events on reconnect, an inline warning banner during reconnection, and a system message after max retries are exhausted. Clean, well-structured change overall — but there's one correctness bug in the skip-counting logic.

---

### 🔴 Critical

**Event counter overcounts, causing event loss on second reconnection**
`AssistantChatPanel.tsx` — `withSkipGuard` (lines 64–73)

`eventsReceivedRef` is incremented for **every** event, including skipped (replayed) ones. This means after a reconnect cycle, the counter includes both the skipped replays and the genuinely new events. On a *second* disconnect, `eventsToSkipRef` is set to this inflated count, which is larger than the number of events the backend will actually replay — causing new, never-seen events to be silently dropped.

**Walkthrough:**
1. Initial connection: 5 events received → `eventsReceivedRef = 5`
2. Disconnect → `eventsToSkipRef = 5`
3. Reconnect: backend replays 5 old events (all skipped, counter still incremented) + 3 new → `eventsReceivedRef = 13`
4. Second disconnect → `eventsToSkipRef = 13`
5. Second reconnect: backend replays 8 events (5 + 3), but skip guard tries to skip 13 → **the next 5 genuine new events are silently dropped**

**Fix:** only increment `eventsReceivedRef` for events that are actually processed (not skipped):

```tsx
function withSkipGuard(handler: (e: MessageEvent) => void): (e: MessageEvent) => void {
    return (e: MessageEvent) => {
        if (eventsToSkipRef.current > 0) {
            eventsToSkipRef.current--;
            return;
        }
        eventsReceivedRef.current++;  // ← move AFTER the skip check
        handler(e);
    };
}
```

This keeps `eventsReceivedRef` in sync with the number of unique events the backend will replay.

---

### 💡 Suggestion

**Consider disabling user input during reconnection**
`AssistantChatPanel.tsx` — render (lines 254–257)

While `isReconnecting` is `true`, the `AssistantMessageInput` is still enabled (it's only disabled by `isProcessing`). A user could type and send a message while the SSE connection is down. The HTTP POST to send the message might succeed, but the response events would be lost during the skip phase or missed entirely. Consider also disabling input while reconnecting:

```tsx
<AssistantMessageInput onSend={handleSend} disabled={isProcessing || isReconnecting} />
```

---

### 💬 Nitpick

**Reconnecting alert placement**
The `Alert` is rendered between `AssistantMessageList` and `AssistantMessageInput`. This works fine visually, but depending on how tall the alert renders, it may cause a slight layout shift that could be jarring. No action needed — just something to verify during manual testing.

---

### Overall Assessment

The architecture of the reconnection logic is sound — exponential backoff, proper cleanup of timers, reset on success, and the `withSkipGuard` abstraction is a nice pattern. The **critical bug** with the event counter will cause real problems for users who experience more than one disconnect in a session, so that should be fixed before merging. With that one-line move, this looks good to go. 👍

PR #105, opened 2026-07-22
Add user-visible error feedback for failed AI assistant operations

EricWittmann:
## Summary

Adds user-visible error feedback for failed operations in the AI Assistant frontend. Previously, most `catch` blocks only called `console.error`, leaving users unaware when actions failed.

### `AssistantSessionPage.tsx`
- Added `endSessionError` state and a dismissible inline `Alert` banner (using PatternFly `AlertActionCloseButton`) that appears when ending a session fails. This follows the existing `applyError` pattern already used for the "Apply All" action.

### `AssistantChatPanel.tsx`
- Added system messages (via the existing `addMessage({ type: "system", ... })` pattern) when:
  - **Sending a message** fails — displays "Failed to send message. Please try again."
  - **Responding to a permission request** fails — displays "Failed to submit permission response. Please try again."
- This reuses the same `type: "system"` message pattern already used for `session_error` SSE events.

## Approach

The fix uses two existing patterns already established in the codebase:
1. **Inline Alert banners** (for page-level actions like ending a session) — matches the `applyError` pattern
2. **System chat messages** (for chat-level actions like sending messages) — matches the `session_error` event handler pattern

Fixes #77

PR #115, opened 2026-07-23
Make operational config properties runtime-configurable via database

### LINKED ISSUE #111: Make operational config properties runtime-configurable via database
## Summary

Several operational tuning knobs (max-turns, timeouts, budgets, confidence thresholds, model
overrides) are currently injected via `@ConfigProperty` at startup and require an application
restart (or environment variable change) to modify. These should be stored in the database and
editable at runtime through the REST API and UI.

## Current State

All `axiom.*` properties are static `@ConfigProperty` values in `application.properties`. The
only runtime-configurable settings today are the Manager's system prompt and prompt template,
stored in the `manager_config` table and exposed via `GET/PUT /api/v1/manager/config`.

### Manager
| Property | Default | Location |
|----------|---------|----------|
| `axiom.manager.max-turns` | `5` | `ManagerService.java` |
| `axiom.manager.confidence-threshold` | `0.7` | `ManagerService.java` |
| `axiom.manager.timeout-seconds` | `120` | `ManagerService.java` |
| `axiom.manager.model` | *(none)* | `ManagerService.java` |

### Claude Code Actor
| Property | Default | Location |
|----------|---------|----------|
| `axiom.claude-code.max-turns` | `50` | `ClaudeCodeActor.java` |
| `axiom.claude-code.max-budget-usd` | `5.0` | `ClaudeCodeActor.java` |
| `axiom.claude-code.timeout-seconds` | `600` | `ClaudeCodeActor.java` |
| `axiom.claude-code.model` | *(none)* | `ClaudeCodeActor.java` |
| `axiom.claude-code.available-models` | *(list)* | `SystemResourceImpl.java` |

### OpenCode Actor
| Property | Default | Location |
|----------|---------|----------|
| `axiom.opencode.max-steps` | `50` | `OpenCodeActor.java` |
| `axiom.opencode.timeout-seconds` | `600` | `OpenCodeActor.java` |
| `axiom.opencode.model` | *(none)* | `OpenCodeActor.java` |
| `axiom.opencode.available-models` | *(list)* | `SystemResourceImpl.java` |

### Assistant
| Property | Default | Location |
|----------|---------|----------|
| `axiom.assistant.max-sessions` | `12` | `AssistantSessionManager.java` |

### Other
| Property | Default | Location |
|----------|---------|----------|
| `axiom.ai-engine` | `claude-code` | `AiEngineProducer.java` |
| `axiom.event-source-logs.retention-days` | `7` | `EventSourceLogCleanup.java` |
| `axiom.script.timeout-seconds` | `60` | `ScriptExecutionService.java` |

### Backend

1. **New entity**: Create a `SystemSettingsEntity` (single-row table, similar pattern to
   `ManagerConfigEntity`) with typed columns for each configurable property.

2. **Flyway migration**: Add a `V30__system_settings.sql` migration to create the table and
   seed it with the current defaults from `application.properties`.

3. **Settings service**: Create a `SystemSettingsService` that loads the single row and
   provides typed getters. Cache the row in memory and invalidate on updates.

4. **Consumers read from service**: Replace the `@ConfigProperty` injections in
   `ManagerService`, `ClaudeCodeActor`, `OpenCodeActor`, `AssistantSessionManager`, etc. with
   calls to `SystemSettingsService`. The `@ConfigProperty` values become fallback defaults
   only (used if the DB row doesn't exist yet, e.g. during first startup before migration).

5. **REST API**: Add `GET/PUT /api/v1/system/settings` endpoints to the OpenAPI spec, with a
   `SystemSettings` schema. The PUT endpoint updates the DB row and invalidates the cache.

### Frontend

6. **Settings UI page**: Add a "Settings" page under the existing "Configuration" sidebar
   section (next to "AI Engine" and "Manager"). The page should group settings by category
   (Manager, Claude Code, OpenCode, Assistant, General) with appropriate input controls
   (number inputs for turns/timeouts/budgets, text inputs for models, dropdowns for engine
   selection).

### Properties NOT to include

These infrastructure-level properties should remain static (env vars / `application.properties`
only):
- `axiom.claude-code.executable` — filesystem path to the CLI binary
- `axiom.opencode.server.hostname` / `axiom.opencode.server.port` — network config
- `axiom.workspace.root` — filesystem path
- Poll interval properties — tied to Quarkus `@Scheduled` annotations

## Considerations

- **Fallback chain**: DB value > `application.properties` value > hardcoded default. If the DB
  row is missing or a column is null, fall back to the `@ConfigProperty` value.
- **Validation**: Enforce reasonable bounds (e.g. max-turns >= 1, timeout > 0, budget >= 0).
- **No restart required**: Changes should take effect on the next task/evaluation execution
  without requiring an application restart.
- **Consider merging Manager prompts**: The existing `manager_config` table could potentially be
  folded into this new `system_settings` table, or kept separate if the prompt fields are
  better managed independently.

EricWittmann:
## Summary

- Move 18 operational tuning knobs from static `@ConfigProperty` injections into a new single-row
  `system_settings` database table, editable at runtime via REST API and UI
- Add `SystemSettingsService` with in-memory cache, typed getters, and `@ConfigProperty` fallback
  for graceful first-startup behavior
- Add `GET/PUT /api/v1/system/settings` endpoints with input validation
- Migrate 13 consumer classes to read settings from the service instead of static injection
- Add a new **Settings** page in the UI with tabbed form sections (General, Manager, Claude Code,
  OpenCode, Assistant)

Infrastructure-level properties (`executable`, `server.hostname/port`, `workspace.root`, poll
intervals) remain static `@ConfigProperty` only.

Closes #111

## Test plan

- [ ] `mvn install` builds successfully (verified)
- [ ] Start app — confirm `system_settings` table is created with one seeded row
- [ ] `GET /api/v1/system/settings` returns all default values
- [ ] `PUT /api/v1/system/settings` with modified values persists and returns updated values
- [ ] Validation rejects invalid values (e.g. negative timeouts, confidence > 1)
- [ ] Change a setting via PUT (e.g. `managerMaxTurns` → 3), trigger a manager evaluation,
  confirm it respects the new value without restart
- [ ] Navigate to Settings page in UI, verify all fields display current values
- [ ] Modify settings in UI, save, reload page — confirm persistence
- [ ] Delete the `system_settings` row from DB, confirm app still works using fallback defaults

### DIFF
--- a/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeActor.java
+++ b/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeActor.java
@@ -4,15 +4,15 @@ import io.apitomy.axiom.actors.spi.Actor;
 import io.apitomy.axiom.actors.spi.ActorContext;
 import io.apitomy.axiom.actors.spi.TaskResult;
 import io.apitomy.axiom.core.entities.TaskEntity;
+import io.apitomy.axiom.core.services.SystemSettingsService;
 import jakarta.enterprise.context.ApplicationScoped;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
+import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -26,17 +26,8 @@ public class ClaudeCodeActor implements Actor {
 
     private static final Logger LOG = Logger.getLogger(ClaudeCodeActor.class);
 
-    @ConfigProperty(name = "axiom.claude-code.model")
-    Optional<String> model;
-
-    @ConfigProperty(name = "axiom.claude-code.max-turns", defaultValue = "50")
-    int maxTurns;
-
-    @ConfigProperty(name = "axiom.claude-code.max-budget-usd", defaultValue = "5.0")
-    double maxBudgetUsd;
-
-    @ConfigProperty(name = "axiom.claude-code.timeout-seconds", defaultValue = "600")
-    int timeoutSeconds;
+    @Inject
+    SystemSettingsService settingsService;
 
     private final Map<Long, ClaudeCodeSubprocess> runningProcesses = new ConcurrentHashMap<>();
 
@@ -67,14 +58,17 @@ public class ClaudeCodeActor implements Actor {
 
         ClaudeCodeCommandBuilder cmdBuilder = ClaudeCodeCommandBuilder
                 .fromContext(prompt, context)
-                .maxTurns(maxTurns)
-                .maxBudgetUsd(maxBudgetUsd);
+                .maxTurns(settingsService.getClaudeCodeMaxTurns())
+                .maxBudgetUsd(settingsService.getClaudeCodeMaxBudgetUsd());
 
         // Per-action-type model takes priority over the global default
         if (context.getModel() != null && !context.getModel().isBlank()) {
             cmdBuilder.model(context.getModel());
         } else {
-            model.ifPresent(cmdBuilder::model);
+            String globalModel = settingsService.getClaudeCodeModel();
+            if (globalModel != null) {
+                cmdBuilder.model(globalModel);
+            }
         }
 
         if (context.getMcpConfigFile() != null) {
@@ -90,7 +84,7 @@ public class ClaudeCodeActor implements Actor {
                 command,
                 workDir,
                 context.getEnvironment() != null ? context.getEnvironment() : Map.of(),
-                Duration.ofSeconds(timeoutSeconds),
+                Duration.ofSeconds(settingsService.getClaudeCodeTimeoutSeconds()),
                 line -> LOG.tracef("Task %d stream: %s", task.id, line),
                 logBuilder
         );
--- a/app/src/main/java/io/apitomy/axiom/app/ActionTypeAiService.java
+++ b/app/src/main/java/io/apitomy/axiom/app/ActionTypeAiService.java
@@ -8,10 +8,10 @@ import io.apitomy.axiom.core.entities.AiUsageEntity;
 import io.apitomy.axiom.engine.spi.AiEngine;
 import io.apitomy.axiom.engine.spi.AiEngineConfig;
 import io.apitomy.axiom.engine.spi.AiEngineResult;
+import io.apitomy.axiom.core.services.SystemSettingsService;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 
 import java.time.Instant;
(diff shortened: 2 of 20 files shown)


EricWittmann (commented):
## Code Review Summary

This PR moves 18 operational tuning knobs from static `@ConfigProperty` injections into a runtime-configurable `system_settings` database table, editable via REST API and a new Settings UI page. The architecture is clean: a single-row Panache entity, an `@ApplicationScoped` service with in-memory `volatile` cache and `@ConfigProperty` fallback, validation in the REST layer, and a Flyway migration to seed the table.

---

### Findings

#### ⚠️ Warning

**1. No validation for `aiEngine` field — invalid values could break the system**
`SystemResourceImpl.validateSettings()` validates all numeric fields but does not validate the `aiEngine` string. A user could set it to `"invalid-engine"` via PUT, which would immediately break assistant session creation (which checks `"claude-code".equals(settingsService.getAiEngine())`) and potentially other engine-dependent behavior. Consider validating it against the registered engine types from `AiEngineRegistry.getAvailableEngineTypes()`.

**2. Detached entity cached after transaction close**
In `SystemSettingsService.load()` (line 99) and `update()` (line 116), the `cached` reference is set to a JPA entity that becomes detached when the `@Transactional` method completes. This works correctly today because all fields are eagerly loaded primitives/strings, but it's a subtle assumption. If someone later adds a lazy relationship to `SystemSettingsEntity`, the cached reference would throw `LazyInitializationException` on access. A defensive approach would be to copy values into a plain DTO or detached snapshot, though for now the current approach is pragmatic and acceptable.

**3. UI "Active Engine" field is a free-text input — should be a dropdown**
In `SystemSettingsPage.tsx` (line 115-120), the AI Engine setting is rendered as a plain `TextInput`. Since the valid engine types are already available via the `fetchEngines()` API, this should use a `<FormSelect>` or similar dropdown to prevent typos and invalid values. This pairs with finding #1 above — together they could allow a user to easily set an invalid engine.

#### 💡 Suggestion

**4. No user-visible feedback on save success/failure**
In `SystemSettingsPage.tsx`, `handleSave()` (line 44-49) catches errors via `console.error` and clears the dirty flag on success, but doesn't show any toast/notification to the user. Given that PR #105 recently added user-visible error feedback for assistant operations, it would be consistent to add a success alert or error alert here too.

**5. Redundant `@Transactional` on `updateSystemSettings`**
`SystemResourceImpl.updateSystemSettings()` (line 163) is annotated with `@Transactional`, and it calls `settingsService.update()` which is also `@Transactional`. Since the inner transaction joins the outer one, this works correctly, but the `@Transactional` on `settingsService.update()` is misleading — it suggests it manages its own transaction boundary when in practice it relies on the caller's. Consider documenting this or removing one of the annotations (probably keep it on the REST method only, since the service method is always called within a transaction context).

**6. Consider upper-bound validation for numeric fields**
The validation in `validateSettings()` enforces minimum values (e.g., `>= 1`) but no maximum bounds. A user could set `managerMaxTurns` to `Integer.MAX_VALUE` or `claudeCodeMaxBudgetUsd` to an extremely large number. While not a critical issue, reasonable upper bounds would prevent accidental resource exhaustion.

#### 📝 Nitpick

**7. Hardcoded default model lists will drift from `application.properties`**
The default available models are hardcoded in three places: the `@ConfigProperty` defaults in `SystemSettingsService`, the SQL migration's `INSERT` statement, and implicitly in the UI. If these lists need updating, all three locations must be changed in sync. This is an inherent trade-off of the fallback design and isn't easily fixed, but worth noting as future maintenance overhead.

---

### Overall Assessment

**Looks good to merge with minor changes.** The architecture is well-designed — the volatile cache with `@ConfigProperty` fallback is a pragmatic approach for a single-server application. The migration correctly seeds the table, the service cleanly encapsulates the caching logic, and the consumer classes are consistently migrated to use the service.

The most actionable finding is **#1 + #3** (validate the engine field and use a dropdown in the UI) since an invalid engine value could break core functionality. The remaining findings are enhancements that could be addressed in a follow-up.

EricWittmann:
Rejecting this change.  I'm going to re-think some of these things.

PR #117, opened 2026-07-23
Handle conversation_reset event from /clear slash command

### LINKED ISSUE #106: Implement support for "/clear" slash command in AI Assistant sessions
## Summary

When a user sends the `/clear` command in an interactive AI Assistant session, the UI currently
displays a warning message:

> Unhandled event type: conversation_reset

The `/clear` command is a built-in Claude Code slash command that resets the conversation. The
backend event parser (`AssistantEventParser`) does not have a case for the `conversation_reset`
event type, so it falls through to the `default` branch and emits an `unhandled_event` SSE event.
The frontend (`AssistantChatPanel.tsx`) then renders this as a warning.

## Expected Behavior

When `/clear` is sent:

1. The backend should recognize the `conversation_reset` event type and emit a semantic
   `conversation_reset` SSE event (or similar).
2. The frontend should handle this event by clearing the conversation message history and
   displaying a brief system message (e.g., "Conversation cleared.").
3. The UI should be ready for the next turn with a fresh conversation state.

## Current Behavior

The `AssistantEventParser.parse()` method's `switch` on event type does not include a
`"conversation_reset"` case. It falls through to `default`, which wraps the raw event as
`unhandled_event`. The frontend displays:

```
Unhandled event type: conversation_reset
```

### Backend (`AssistantEventParser.java`)

Add a new case in the `switch` statement in `parse()`:

```java
case "conversation_reset" -> parseConversationReset(node);
```

The `parseConversationReset` method should emit a semantic event (e.g., `conversation_reset`)
so the frontend can act on it.

### Frontend (`AssistantChatPanel.tsx`)

Add a handler for the new `conversation_reset` event type:

- Clear the messages list
- Reset processing state
- Optionally show a system message like "Conversation cleared."

## Files Likely Affected

- `app/src/main/java/io/apitomy/axiom/app/assistant/AssistantEventParser.java`
- `app/src/test/java/io/apitomy/axiom/app/assistant/AssistantEventParserTest.java`
- `ui/src/components/assistant/AssistantChatPanel.tsx`

EricWittmann:
## Summary

- Add `conversation_reset` case to `AssistantEventParser.parse()` so the backend emits a
  semantic event instead of falling through to `unhandled_event`
- Clear in-memory `eventHistory` in `AssistantSession.readStdout()` when a reset occurs,
  so page refreshes and SSE reconnects see a clean conversation
- Handle `conversation_reset` in `AssistantChatPanel.tsx` by replacing messages with a
  "Conversation cleared." system message

Resolves #106

## Test plan

- [ ] `mvn install` passes (includes new `parseConversationResetEvent` unit test)
- [ ] Start an assistant session, send messages, type `/clear` — messages clear with a
  system message, no "Unhandled event type" warning
- [ ] Refresh the page after `/clear` — only the "Conversation cleared." message appears
- [ ] Continue chatting after `/clear` — new messages appear normally

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantEventParser.java
+++ b/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantEventParser.java
@@ -55,6 +55,7 @@ public class AssistantEventParser {
                 case "result" -> parseResult(node);
                 case "sdk_control_request" -> parseSdkControlRequest(node);
                 case "control_request" -> parseControlRequest(node);
+                case "conversation_reset" -> parseConversationReset(node);
                 default -> {
                     ObjectNode data = JsonNodeFactory.instance.objectNode();
                     data.put("rawType", type);
@@ -180,6 +181,18 @@ public class AssistantEventParser {
         data.put("description", request.path("description").asText());
         data.set("toolInput", request.path("input"));
         return List.of(new SseEvent("permission_request", data));
+    }
+
+    /**
+     * Parses a conversation_reset event emitted when the user sends the
+     * {@code /clear} slash command.
+     *
+     * @param root the raw NDJSON node
+     * @return a single conversation_reset SSE event
+     */
+    private List<SseEvent> parseConversationReset(JsonNode root) {
+        ObjectNode data = JsonNodeFactory.instance.objectNode();
+        return List.of(new SseEvent("conversation_reset", data));
     }
 
     /**
--- a/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantEventParser.java
+++ b/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantEventParser.java
@@ -25,6 +25,7 @@ import java.util.List;
  *   <li>{@code user} (with tool_use_result) → {@code tool_result}</li>
  *   <li>{@code result} → {@code turn_complete}</li>
  *   <li>{@code sdk_control_request} / {@code control_request} → {@code permission_request}</li>
+ *   <li>{@code conversation_reset} → {@code conversation_reset}</li>
  * </ul>
  */
 public class AssistantEventParser {
@@ -199,7 +200,8 @@ public class AssistantEventParser {
      * A normalised SSE event parsed from Claude Code's NDJSON output.
      *
      * @param type the normalised event type (session_init, assistant_text,
-     *             tool_use, tool_result, turn_complete, permission_request, thinking)
+     *             tool_use, tool_result, turn_complete, permission_request,
+     *             thinking, conversation_reset)
      * @param data the extracted/transformed JSON data
      */
     public record SseEvent(String type, JsonNode data) {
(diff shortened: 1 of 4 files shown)


EricWittmann (commented):
## Code Review

This PR adds end-to-end handling for the `conversation_reset` event emitted by Claude Code when the user types `/clear`. It touches the NDJSON parser (new case + method), session history management (clearing event history), and the frontend chat panel (replacing messages with a system notification). The code is well-structured and follows existing patterns. However, there is a significant event index interaction bug.

---

### 🔴 Critical

**Event index reset causes `conversation_reset` to be silently dropped by the live SSE listener**

In `AssistantSession.readStdout()`, the `eventHistory.clear()` resets the list to empty, then `eventHistory.add(event)` adds the `conversation_reset` event at position 0. The CDI bridge listener in `AssistantSessionManager` (line ~226) calculates:

```java
int eventIndex = session.getEventCount() - 1; // → 0 after clear + add
```

The frontend's SSE callback in `AssistantChatPanel.tsx` (line ~245) guards against duplicates:

```typescript
if (eventIndex <= lastSeenIndexRef.current) {
    // Skipping (already seen)
    return;
}
```

After a conversation with, say, 50 prior events, `lastSeenIndexRef.current` will be ~50. The `conversation_reset` arrives with `eventIndex = 0`, which is `<= 50`, so **it gets skipped**. The client never calls `processEvent("conversation_reset", ...)`, the messages are never cleared, and the `/clear` command appears to do nothing.

**Impact:** The `/clear` slash command will not work for the currently-connected client. A page refresh would work (since `lastSeenIndexRef` resets to -1), but that defeats the purpose.

**Suggested fix options:**
1. **Don't clear `eventHistory` in `readStdout`** — just append the `conversation_reset` event. The frontend already handles it by replacing all messages, so the live path works. For page refresh/reconnect, the history replay would process old events and then the `conversation_reset` would clear them, resulting in the correct final state (albeit with a brief flash of old messages).
2. **Use a monotonically increasing counter** for `eventIndex` instead of deriving it from `eventHistory.size()`. This way, even after clearing the list, the index continues from where it left off and the frontend won't skip the event.

---

### ⚠️ Warning

**`addListenerWithHistory` reconnect semantics are broken after clear**

Related to the above: `addListenerWithHistory(listener, sinceId)` computes:

```java
int fromIndex = (int) Math.min(sinceId + 1, eventHistory.size());
```

After a clear, if `sinceId` was 50 and `eventHistory.size()` is now 2, `fromIndex` clamps to 2 and `subList(2, 2)` is empty — the client misses all post-clear events during the replay. Any SSE reconnect after a `/clear` would leave the UI stale until a full page refresh.

This is a secondary consequence of the same root cause and would be resolved by either of the fix options above.

---

### ✅ What looks good

- **Parser change** — Clean addition following the existing `switch` pattern. The `parseConversationReset` method is consistent with other parse methods.
- **Test coverage** — The new `parseConversationResetEvent` test follows the established style and verifies the essential behavior.
- **Frontend handler** — The `conversation_reset` case in `processEvent` correctly replaces all messages with a system notification and resets the processing state. This is the right UX.
- **Javadoc** — Well-documented method with proper `@param` / `@return` tags.

---

### Overall Assessment

The PR has a **critical bug** that will prevent the `/clear` command from working for the live-connected client. The event index resetting to 0 after `eventHistory.clear()` causes the frontend's deduplication guard to silently drop the event. I'd recommend addressing this before merging — either by not clearing the history (option 1, minimal change) or by introducing a monotonic counter (option 2, more robust but larger change).

PR #122, opened 2026-07-27
fix: subprocess join timeout + pre-filter bot events

### LINKED ISSUE #121: fix: ClaudeCodeSubprocess.join() hangs when orphan processes keep pipes open
## Bug

When a Claude Code subprocess exits with code 1 (e.g., Vertex AI auth failure or rate limit), `ClaudeCodeSubprocess.doExecute()` can hang indefinitely at `stdoutFuture.join()` / `stderrFuture.join()`.

## Root Cause

In `ClaudeCodeSubprocess.java`, after `process.waitFor(timeout)` succeeds (process exits), the code calls:

```java
stdoutFuture.join();
stderrFuture.join();
```

These `join()` calls have **no timeout**. If the Claude process spawned child processes that inherit stdout/stderr file descriptors, the reader threads (`readLine()`) never see EOF even though the parent process exited. The `join()` then blocks forever.

This bypasses the 600-second timeout entirely — observed tasks hanging for 65+ minutes with 0 tokens and `$n/a` cost.

## Evidence

From production logs:

```
19:32:51 ClaudeCodeActor    Executing task 91 (action: review-pr) via Claude Code
19:32:51 ClaudeCodeSubprocess Launching Claude Code
20:37:49 TaskExecutionService Task 91 failed (cost: $n/a, tokens: 0/0)
```

65 minutes elapsed despite a 600-second timeout. The process likely exited with code 1 early, but `join()` hung on the reader futures.

## Fix

Add a timeout to the `join()` calls:

```java
try {
    stdoutFuture.get(30, TimeUnit.SECONDS);
} catch (TimeoutException e) {
    stdoutFuture.cancel(true);
    LOG.warnf("Stdout reader did not finish within 30s after process exit");
}
try {
    stderrFuture.get(30, TimeUnit.SECONDS);
} catch (TimeoutException e) {
    stderrFuture.cancel(true);
    LOG.warnf("Stderr reader did not finish within 30s after process exit");
}
```

Or use `Process.toHandle().descendants().forEach(ProcessHandle::destroyForcibly)` before reading, to kill any orphan child processes that might hold the pipes open.

## Impact

Affects any deployment using Vertex AI or other auth backends that can return exit code 1. With many concurrent events (164 Claude launches/day observed), rate limits trigger frequently, and each hung task blocks an actor slot for 60+ minutes instead of failing fast.

carlesarnal:
## Summary

Two fixes for event processing efficiency:

1. **Subprocess join timeout** — `stdoutFuture.join()` / `stderrFuture.join()` in `ClaudeCodeSubprocess` can hang indefinitely when orphan child processes keep pipes open. Adds 30-second bounded wait and kills the full process tree on exit.

2. **Pre-filter bot events** — Events from known bots (`github-actions[bot]`, `sonarqubecloud[bot]`, `renovate[bot]`, `dependabot[bot]`, `codecov[bot]`) and slash-command comments (`/ready`, `/retry`, `/accept`, etc.) are skipped before invoking the AI Manager. These were 50%+ of all events and were always ignored after an expensive Manager evaluation.

## Impact

On a repo with 225 events over 3 days:
- 74 bot comments (43 github-actions, 31 sonarqubecloud) → would be pre-filtered
- 10+ slash commands (/retry, /accept) → would be pre-filtered
- Saves ~$5 per 225 events in Manager evaluation costs
- Subprocess fix prevents 65-minute actor slot blockage on auth failures

## Test plan

- [x] `mvn test -pl actors/claude-code` — 36 tests pass
- [x] `mvn test-compile -pl app -am -DskipTests` compiles cleanly
- [ ] Verify pre-filtered events logged as "event-pre-filtered" in activity log
- [ ] Verify bot events no longer reach the Manager

Closes #121

### DIFF
--- a/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeSubprocess.java
+++ b/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeSubprocess.java
@@ -87,7 +87,7 @@ public class ClaudeCodeSubprocess {
     public void kill() {
         if (process != null && process.isAlive()) {
             LOG.warn("Killing Claude Code subprocess");
-            process.destroyForcibly();
+            destroyProcessTree();
         }
     }
 
--- a/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeSubprocess.java
+++ b/actors/claude-code/src/main/java/io/apitomy/axiom/actors/claudecode/ClaudeCodeSubprocess.java
@@ -13,7 +13,9 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 
 /**
@@ -164,17 +166,19 @@ public class ClaudeCodeSubprocess {
         boolean completed = process.waitFor(timeout.toMillis(), TimeUnit.MILLISECONDS);
         if (!completed) {
             LOG.warnf("Claude Code subprocess timed out after %s", timeout);
-            process.destroyForcibly();
-            process.waitFor(5, TimeUnit.SECONDS);
+            destroyProcessTree();
             logBuilder.footer("Timed Out", null, null, null,
                     Duration.between(startTime, Instant.now()));
             return new ClaudeCodeResult("Process timed out after " + timeout,
                     null, null, null, null, 124, logBuilder.build());
         }
 
-        // Wait for stream readers to finish
-        stdoutFuture.join();
-        stderrFuture.join();
+        // Kill any orphan child processes that may hold pipes open
+        destroyProcessTree();
+
+        // Wait for stream readers to finish (bounded to avoid hanging on orphan pipes)
+        joinWithTimeout(stdoutFuture, "stdout");
+        joinWithTimeout(stderrFuture, "stderr");
 
         int exitCode = process.exitValue();
         LOG.tracef("Claude Code subprocess exited with code %d", exitCode);
@@ -254,6 +258,33 @@ public class ClaudeCodeSubprocess {
                     jsonLine.substring(0, Math.min(jsonLine.length(), 200)));
             // Fall back to treating the raw output as the result text
             return new ClaudeCodeResult(jsonLine, null, null, null, null, exitCode, null);
+        }
+    }
+
+    private void destroyProcessTree() {
+        process.descendants().forEach(ph -> {
+            LOG.tracef("Killing descendant process %d", ph.pid());
+            ph.destroyForcibly();
+        });
+        process.destroyForcibly();
+        try {
+            process.waitFor(5, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private void joinWithTimeout(CompletableFuture<Void> future, String name) {
+        try {
+            future.get(30, TimeUnit.SECONDS);
+        } catch (TimeoutException e) {
+            LOG.warnf("%s reader did not finish within 30s after process exit, cancelling", name);
+            future.cancel(true);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            future.cancel(true);
+        } catch (ExecutionException e) {
+            LOG.warnf(e.getCause(), "Error in %s reader", name);
         }
     }
 
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -41,6 +41,11 @@ public class PipelineOrchestrator {
 
     private static final Logger LOG = Logger.getLogger(PipelineOrchestrator.class);
 
+    private static final Set<String> SLASH_COMMANDS = Set.of(
+            "/ready", "/retry", "/accept", "/reject", "/disable-tests",
+            "/enable-tests", "/unstale", "/skip-review", "/auto-merge", "/merge"
+    );
+
     @Inject
     ManagerService managerService;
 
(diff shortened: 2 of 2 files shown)


EricWittmann (commented):
## Code Review

This PR fixes a production hang caused by orphan child processes keeping stdout/stderr pipes open after the main Claude Code subprocess exits. It adds a `destroyProcessTree()` method to kill all descendant processes and a `joinWithTimeout()` method to bound the reader future waits to 30 seconds.

---

### Findings

#### ⚠️ Warning

**`CompletableFuture.cancel(true)` does not interrupt the running thread**
`ClaudeCodeSubprocess.java` — `joinWithTimeout()`

```java
future.cancel(true);
```

`CompletableFuture.cancel(true)` ignores the `mayInterruptIfRunning` flag — it only marks the future as cancelled but does **not** interrupt the thread blocked on `readLine()`. This means if the 30-second timeout fires, the reader thread will continue to block until the pipe is eventually closed by other means.

In practice, this is mitigated because `destroyProcessTree()` is called *before* `joinWithTimeout()`, so pipes should already be closing by the time the timeout fires. The `cancel()` call still serves a purpose — it prevents the future from being observed as "completed normally" — but it won't actually stop a stuck reader thread.

If you want true thread interruption as a last resort, you could use `ExecutorService.submit()` instead of `CompletableFuture.runAsync()` and hold a reference to the `Future` returned by `submit()`, which does honor `cancel(true)`. However, given the belt-and-suspenders approach here (kill process tree first, then wait), this is likely fine as-is. Just worth being aware of.

#### 💡 Suggestion

**`kill()` method doesn't kill descendant processes**
`ClaudeCodeSubprocess.java`, line 88–91

```java
public void kill() {
    if (process != null && process.isAlive()) {
        LOG.warn("Killing Claude Code subprocess");
        process.destroyForcibly();
    }
}
```

The existing `kill()` method (used for external cancellation) still calls only `process.destroyForcibly()` without killing descendants. This could leave the same orphan child processes that this PR is fixing in the `doExecute()` path. Consider updating `kill()` to also use `destroyProcessTree()` for consistency:

```java
public void kill() {
    if (process != null && process.isAlive()) {
        LOG.warn("Killing Claude Code subprocess");
        destroyProcessTree();
    }
}
```

This is outside the scope of the current fix but would prevent the same class of orphan-process issues when tasks are externally cancelled.

#### 💡 Suggestion

**Extract the 30-second timeout as a named constant**
`ClaudeCodeSubprocess.java`, line 279

```java
future.get(30, TimeUnit.SECONDS);
```

Consider extracting this as a constant for clarity and easier tuning:

```java
private static final Duration READER_JOIN_TIMEOUT = Duration.ofSeconds(30);
```

#### 🔍 Nitpick

**Timeout path still doesn't join reader futures**
`ClaudeCodeSubprocess.java`, lines 167–173

In the timeout path, after `destroyProcessTree()`, the method returns immediately without joining the reader futures. This is pre-existing behavior (not a regression), but the reader threads could briefly outlive the returned result. Since `destroyProcessTree()` closes the pipes, they should terminate quickly, so this is low-risk. Just noting for awareness.

---

### Overall Assessment

**Looks good to merge.** ✅

This is a well-targeted fix for a real production issue. The approach is sound: kill the process tree to close pipes, then apply a bounded wait as a safety net. The code is clean, the error handling is thorough (proper `InterruptedException` propagation, `ExecutionException` logging), and the PR description clearly explains the problem, root cause, and fix.

The warning about `CompletableFuture.cancel(true)` is worth knowing about but isn't a practical concern given the overall design. The suggestion about `kill()` is worth considering as a follow-up.

EricWittmann (commented):
## Code Review Summary

This PR addresses two related issues: (1) preventing indefinite hangs from orphan child processes keeping subprocess pipes open after `ClaudeCodeSubprocess` completes, and (2) pre-filtering known bot events and slash commands before they reach the expensive AI Manager evaluation. Both changes are well-motivated with clear problem descriptions and measurable impact.

---

### Findings

#### ⚠️ Warning

**Misplaced Javadoc — `PipelineOrchestrator.java` (lines 622-674)**

The `BOT_LOGINS` / `SLASH_COMMANDS` constants and the `shouldSkipEvent()` method were inserted between the existing Javadoc comment for `completeTraceFailed` and the method itself. The result:

```java
/**
 * Completes the trace with a failed status on pipeline error.   // ← this Javadoc...
 */
private static final Set<String> BOT_LOGINS = Set.of(              // ← ...now documents this field
        "github-actions[bot]", ...
);

// ... 50 lines of new code ...

private void completeTraceFailed(TraceContext traceCtx) {           // ← method is orphaned from its doc
```

Move the constants and `shouldSkipEvent()` method above the `completeTraceIfSync` / `completeTraceFailed` block (or to the top of the class with other constants) so the Javadoc association is preserved.

---

#### 💡 Suggestion

**`kill()` method should use `destroyProcessTree()`  — `ClaudeCodeSubprocess.java` (line 87-91)**

The new `destroyProcessTree()` method properly kills all descendant processes, but the existing public `kill()` method still only calls `process.destroyForcibly()`. If an external caller cancels a running subprocess via `kill()`, orphan child processes would survive — the same problem this PR aims to fix. Consider refactoring `kill()` to delegate to `destroyProcessTree()`:

```java
public void kill() {
    if (process != null && process.isAlive()) {
        LOG.warn("Killing Claude Code subprocess");
        destroyProcessTree();
    }
}
```

---

#### 💡 Suggestion

**`CompletableFuture.cancel(true)` does not interrupt the running thread — `ClaudeCodeSubprocess.java` (line 282)**

Unlike `Future.cancel(true)` from `ExecutorService.submit()`, `CompletableFuture.cancel(true)` ignores the `mayInterruptIfRunning` flag and does **not** interrupt the underlying thread. In the `TimeoutException` catch block, the call `future.cancel(true)` will mark the future as cancelled but won't actually stop a reader thread blocked on `readLine()`.

In practice, this is fine here because `destroyProcessTree()` is called first, which should close the pipes and unblock the readers. But it's worth a comment so future readers don't assume the thread is interrupted:

```java
// Note: cancel(true) on CompletableFuture does NOT interrupt the thread;
// we rely on destroyProcessTree() having closed the pipes.
future.cancel(true);
```

---

#### 💡 Suggestion

**Consider making bot/slash-command lists configurable — `PipelineOrchestrator.java`**

The `BOT_LOGINS` and `SLASH_COMMANDS` sets are hardcoded. Different deployments may interact with additional bots (e.g., `mergify[bot]`, `stale[bot]`, `allcontributors[bot]`) or have custom slash commands. Making these configurable via Quarkus `@ConfigProperty` (with sensible defaults) would increase flexibility without much added complexity. Not a blocker, but worth considering.

---

#### 💡 Suggestion

**No tests for `shouldSkipEvent()` — `PipelineOrchestrator.java`**

The pre-filtering logic handles multiple payload structures (comment vs. top-level user, various event types, edge cases like null/blank payloads). Unit tests would help validate correctness and guard against regressions — especially given the variety of GitHub webhook payload shapes. The PR description mentions manual test plan items but no automated coverage for the new logic.

---

#### 🔍 Nitpick

**30-second timeout is a magic number — `ClaudeCodeSubprocess.java` (line 279)**

The 30-second join timeout could be extracted into a named constant (e.g., `READER_JOIN_TIMEOUT_SECONDS`) for clarity and easier tuning, similar to how `timeout` is already a configurable parameter.

---

### Overall Assessment

**Looks good — minor changes recommended.** The core fixes are solid and well-reasoned. The subprocess process tree cleanup and bounded joins directly address the 65-minute hang scenario. The pre-filtering optimization is a pragmatic cost-saver with clean fallback behavior (parse failures fall through to normal processing).

The main item to address before merge is the **misplaced Javadoc** which will cause confusion for anyone reading the code. The `kill()` method consistency is also worth fixing while you're in this area. The rest are suggestions for consideration.

EricWittmann:
I think the **Pre-filter bot events** feature is a good one.  However, I have long been considering a general purpose filtering layer added to the Eventing feature of Axiom.  Something that would give users the ability to pre-filter the stream of events, allowing only the events they care about through to the Manager.  wdyt?

carlesarnal:
> I think the **Pre-filter bot events** feature is a good one. However, I have long been considering a general purpose filtering layer added to the Eventing feature of Axiom. Something that would give users the ability to pre-filter the stream of events, allowing only the events they care about through to the Manager. wdyt?

I think it probably makes more sense, I wanted something quick because I've been bitten by the rate limits.

PR #123, opened 2026-07-27
Auto-include MCP servers implied by session template allowed tools

EricWittmann:
## Summary
- Custom tools listed in a session template's Allowed Tools were not callable from interactive AI
  assistant sessions because the corresponding MCP servers (e.g. `@axiom-tools`) were not
  automatically started — users had to manually add them to the MCP Servers list as well.
- The task execution path (`McpConfigGenerator.generateMcpConfig`) already derives which MCP servers
  to start from the allowed tools list; this change brings the assistant session path into parity.
- Adds `augmentServersFromAllowedTools()` to auto-include `@axiom-tools`, `@axiom-sdk`, and external
  MCP servers when their tools appear in the resolved allowed tools list.

## Test plan
- [ ] Create a session template with custom tools in Allowed Tools but no `@axiom-tools` in MCP
      Servers — verify the tools are now callable from the interactive session
- [ ] Create a session template with `@axiom-tools` already in MCP Servers — verify no duplicate
      server entries or errors
- [ ] Create a session template with `@axiom-sdk` tools in Allowed Tools — verify the SDK server
      is auto-included
- [ ] Create a session template with an external MCP server tool in Allowed Tools — verify the
      external server is auto-included
- [ ] Verify task execution path still works as before (no regression)

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSessionManager.java
+++ b/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSessionManager.java
@@ -157,12 +157,15 @@ public class AssistantSessionManager {
             Path sessionDir = contextBuilder.createSessionDirectory(sessionId, null);
 
             try {
-            // Resolve MCP servers from template
-            Map<String, AssistantContextBuilder.McpServerConfig> mcpConfigs =
-                    resolveMcpServers(template, project, sessionDir, resolvedEnv);
-
-            // Resolve allowed tools from template
+            // Resolve allowed tools first — MCP server resolution needs them
+            // to auto-include servers implied by tool names.
             List<String> resolvedAllowedTools = resolveAllowedTools(template);
+
+            // Resolve MCP servers from template, augmented with any servers
+            // implied by the allowed tools list.
+            Map<String, AssistantContextBuilder.McpServerConfig> mcpConfigs =
+                    resolveMcpServers(template, project, sessionDir, resolvedEnv,
+                            resolvedAllowedTools);
 
             // Build and write MCP config JSON to session directory
             String mcpConfig = contextBuilder.buildMcpConfig(mcpConfigs);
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review Summary

This PR fixes a parity gap between the task execution path and the interactive assistant session path. Previously, users had to manually add MCP servers (e.g. `@axiom-tools`) to a session template's MCP Servers list even when the corresponding tools were already listed in Allowed Tools. The change reorders resolution so that allowed tools are resolved first, then a new `augmentServersFromAllowedTools()` method auto-includes `@axiom-tools`, `@axiom-sdk`, and external MCP servers whose tools appear in the resolved allowed tools list — mirroring the existing logic in `McpConfigGenerator.generateMcpConfig`.

---

### Findings

#### 💡 Suggestion

**Duplicated prefix constants — consider sharing with `McpConfigGenerator`**
`augmentServersFromAllowedTools` uses inline string literals `"mcp__axiom-tools__"` and `"mcp__axiom-sdk__"` for prefix matching, while `McpConfigGenerator` already defines these as named constants (`AXIOM_TOOLS_PREFIX`, `AXIOM_SDK_PREFIX`). Consider either referencing those existing constants or extracting them to a shared location to keep the prefixes in sync and avoid the risk of a typo divergence.

*File: `AssistantSessionManager.java`, lines 820–822, 836*

#### 💡 Suggestion

**Potential for shared auto-derivation logic**
The `augmentServersFromAllowedTools` logic is a purpose-built mirror of the filtering logic in `McpConfigGenerator.generateMcpConfig` (lines 109–121 and 424–428). While the two paths are structurally different enough that some duplication is acceptable today, if more server types are added in the future, consider extracting a shared utility (e.g. a static helper) that both paths can call. This would prevent them from drifting out of sync.

#### ✅ Positive observations

- **Correct reordering**: Moving `resolveAllowedTools()` before `resolveMcpServers()` is necessary and correctly done. The dependency is clearly documented in comments.
- **Null/empty guard**: `augmentServersFromAllowedTools` short-circuits on null/empty `allowedTools`, preventing unnecessary work and null issues.
- **Duplicate prevention**: The `!serverNames.contains(...)` checks properly prevent duplicate entries when a server is already explicitly listed in the template.
- **External server parsing**: The `mcp__<name>__<tool>` extraction logic (lines 837–847) correctly excludes the built-in prefixes and handles the `secondSep > 5` guard to avoid malformed tool names.
- **Good Javadoc**: The `@link` reference to `McpConfigGenerator#generateMcpConfig` makes the design relationship explicit and discoverable.
- **Informative logging**: The `LOG.infof` calls for each auto-included server are helpful for debugging session setup without being excessively verbose.

---

### Overall Assessment

**Looks good to merge.** The change is well-scoped, correctly mirrors the existing task-execution auto-derivation pattern, and handles edge cases (null/empty lists, duplicates, malformed tool names). The two suggestions above are minor hygiene improvements that could be addressed in a follow-up.

PR #130, opened 2026-07-28
Add "Allow All" toggle to AI Assistant session toolbar

### LINKED ISSUE #109: Add "Allow All" toggle to AI Assistant session toolbar
## Summary

Add a toggle in the AI Assistant session header/toolbar that temporarily enables an "Allow All"
mode for the session. When enabled, all tool permission requests are automatically approved
without requiring user interaction.

## Motivation

During interactive AI Assistant sessions, frequent permission prompts can interrupt workflow — 
especially when the user trusts the tools being invoked and wants to give the assistant free
rein. The existing auto-approval system works at a per-tool/pattern granularity, which is great
for targeted control but cumbersome when the user simply wants to allow everything temporarily.

An "Allow All" toggle provides a quick, session-scoped way to suppress permission prompts
entirely, similar to how Claude Code CLI's `--dangerously-skip-permissions` flag works but with
a more controlled, reversible, in-session UX.

## Proposed Behavior

- **Toggle location**: In the session header/toolbar, near the existing auto-approval shield
  icon. Could be implemented as a toggle switch or a button that visually indicates on/off state.
- **Scope**: Per-session only. The toggle state is transient (lives in the in-memory
  `AssistantSession` object) and resets when the session ends. It is never persisted to the
  database.
- **When enabled**:
  - All incoming `permission_request` events (both `sdk_control_request` and `control_request`
    formats) are automatically approved before reaching the UI.
  - A visual indicator (e.g., warning-colored badge, shield icon change) makes it clear the mode
    is active.
  - Permission events still appear in the chat stream but are shown as auto-approved (not hidden
    entirely, so the user retains visibility into what was allowed).
- **When disabled** (toggled off mid-session):
  - Normal permission prompting resumes immediately.
  - Previously auto-approved permissions are not revoked (they already executed).
- **Safety considerations**:
  - Show a confirmation dialog when enabling, warning the user that all tool invocations will be
    approved automatically.
  - Consider a distinct visual treatment for the session header when "Allow All" is active (e.g.,
    a subtle warning color, similar to the existing plan-mode orange border treatment).

### Backend

- Add an `allowAll` boolean field (default `false`) to `AssistantSession` (in-memory, volatile).
- Add REST endpoints or extend existing ones:
  - `PUT /api/v1/assistant/sessions/{sessionId}/allow-all` with a boolean body to toggle the
    state (update `openapi.json` first per API-first workflow).
  - Alternatively, extend the existing `PATCH /sessions/{sessionId}` endpoint.
- In `AssistantSession.handleAutoApproval()`, check the `allowAll` flag first — if `true`,
  immediately call `respondToPermission(requestId, true, toolInput)` and return `true` before
  evaluating individual auto-approval rules.
- Emit an SSE event (e.g., `allow_all_changed`) when the toggle state changes, so the UI can
  react in real-time across multiple tabs/windows.
- Include the `allowAll` state in `AssistantSessionInfo` so the UI can initialize correctly on
  page load or SSE reconnect.

### Frontend

- Add a toggle component in the session header (`AssistantSessionPage.tsx`), positioned near the
  existing auto-approval shield icon.
- Show a confirmation modal (`PatternFly Modal`) when the user enables the toggle, explaining
  the implications.
- When active, apply a visual indicator to the header (similar to plan-mode styling) so the
  state is obvious at a glance.
- Permission events arriving while "Allow All" is active should render as pre-resolved
  (auto-approved) in the chat stream — no action buttons, just a status indicator.
- Listen for the `allow_all_changed` SSE event to sync state across tabs.

### Existing Infrastructure to Leverage

- The `AutoApprovalRule` / `handleAutoApproval()` pattern in `AssistantSession.java` already
  demonstrates the interception point for auto-approving permissions — "Allow All" is a
  short-circuit at the top of this method.
- The auto-approval shield icon and badge in the toolbar provide a natural visual anchor for the
  new toggle.
- The `permission_resolved` event type is already used to record auto-approvals in event history,
  ensuring SSE replay works correctly.

EricWittmann:
## Summary

Adds a per-session "Allow All" toggle to the AI Assistant session toolbar that auto-approves all
tool permission requests without user interaction. Closes #109.

- **API**: New `PUT /assistant/sessions/{sessionId}/allow-all` endpoint with `SetAllowAllRequest`
  schema; `AssistantSessionInfo` extended with `allowAll` field
- **Backend**: `allowAll` short-circuit in `handleAutoApproval()` that auto-approves permissions
  while keeping them visible in the event stream; emits `allow_all_changed` SSE event for
  cross-tab sync
- **Frontend**: `Switch` toggle in session header with confirmation modal on enable, warning-colored
  visual indicator when active, and SSE-driven state sync across tabs

## Test plan

- [ ] Start a session — verify the Allow All switch is visible and OFF by default
- [ ] Toggle ON — confirmation modal appears with warning text
- [ ] Confirm — switch activates, header shows warning-colored left border
- [ ] Trigger a tool invocation — permission is auto-approved and visible in chat as resolved
- [ ] Toggle OFF — normal permission prompting resumes for subsequent requests
- [ ] Open a second tab/window — verify Allow All state syncs via SSE
- [ ] End session and start a new one — verify Allow All defaults to OFF (not persisted)
- [ ] Toggle on/off rapidly — no errors or stuck state

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java
+++ b/app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java
@@ -70,6 +70,7 @@ public class AssistantSession {
     private final CopyOnWriteArrayList<Consumer<SseEvent>> listeners = new CopyOnWriteArrayList<>();
     private final Object eventLock = new Object();
     private final CopyOnWriteArrayList<AutoApprovalRule> autoApprovalRules = new CopyOnWriteArrayList<>();
+    private volatile boolean allowAll;
     private final Long projectId;
     private final String projectName;
 
@@ -480,6 +481,24 @@ public class AssistantSession {
     }
 
     /**
+     * Returns whether Allow All mode is active for this session.
+     *
+     * @return true if all permissions are auto-approved
+     */
+    public boolean isAllowAll() {
+        return allowAll;
+    }
+
+    /**
+     * Enables or disables Allow All mode for this session.
+     *
+     * @param allowAll true to auto-approve all permissions
+     */
+    public void setAllowAll(boolean allowAll) {
+        this.allowAll = allowAll;
+    }
+
+    /**
      * Checks whether a tool permission request matches any auto-approval rule.
      *
      * @param toolName the tool name from the permission request
@@ -555,10 +574,25 @@ public class AssistantSession {
         }
         String toolName = event.data().path("toolName").asText("");
         JsonNode toolInput = event.data().path("toolInput");
+        String requestId = event.data().path("requestId").asText("");
+
+        // Allow All short-circuit: auto-approve everything but keep the event
+        // visible in the stream so the UI can show it as auto-approved.
+        if (allowAll) {
+            LOG.infof("Auto-approving %s (allow-all) in session %s", toolName, id);
+            try {
+                addEvent(event);
+                respondToPermission(requestId, true, toolInput);
+            } catch (IOException e) {
+                LOG.warnf(e, "Failed to auto-approve %s in session %s", toolName, id);
+                return false;
+            }
+            return true;
+        }
+
         if (!checkAutoApproval(toolName, toolInput)) {
             return false;
         }
-        String requestId = event.data().path("requestId").asText("");
         LOG.infof("Auto-approving %s (rule matched) in session %s", toolName, id);
         try {
             respondToPermission(requestId, true, toolInput);
--- a/app/src/main/java/io/apitomy/axiom/app/rest/AssistantResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/AssistantResourceImpl.java
@@ -516,9 +516,10 @@ public class AssistantResourceImpl implements AssistantResource {
     @Override
     public void setAllowAll(String sessionId, SetAllowAllRequest data) {
         AssistantSession session = requireSession(sessionId);
-        session.setAllowAll(data.getEnabled());
+        boolean enabled = data.getEnabled() != null && data.getEnabled();
+        session.setAllowAll(enabled);
         ObjectNode eventData = objectMapper.createObjectNode();
-        eventData.put("enabled", data.getEnabled());
+        eventData.put("enabled", enabled);
         session.addEvent(new SseEvent("allow_all_changed", eventData));
     }
 
--- a/app/src/main/java/io/apitomy/axiom/app/rest/AssistantResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/AssistantResourceImpl.java
@@ -3,6 +3,7 @@ package io.apitomy.axiom.app.rest;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.apitomy.axiom.api.beans.Environment;
 import io.apitomy.axiom.api.AssistantResource;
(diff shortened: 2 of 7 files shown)


EricWittmann (commented):
## Review Summary

This PR adds a per-session "Allow All" toggle to the AI Assistant toolbar that auto-approves all tool permission requests without user interaction. It spans backend (new REST endpoint, session state, auto-approval logic), API (OpenAPI spec), and frontend (Switch toggle, confirmation modal, SSE sync). The implementation is clean, well-structured, and consistent with existing codebase patterns.

---

### Findings

#### ⚠️ Warning

**1. No user-visible error feedback on Allow All toggle failures**
`ui/src/pages/AssistantSessionPage.tsx` — `handleAllowAllConfirm` / `handleAllowAllDisable`

Both handlers silently swallow API errors with only a `console.error`. Other analogous operations in this page set error state for display (e.g., `setRenameError`, `setEndSessionError`, `setAutoApprovalError`). If the API call fails, the user sees the toggle snap back but gets no explanation.

```tsx
// handleAllowAllConfirm
} catch (err) {
    console.error("Failed to enable Allow All:", err);
    // No setAllowAllError(...) or similar
}
```

Consider adding a visible error notification consistent with the rest of the page.

---

**2. Potential NPE if `SetAllowAllRequest.getEnabled()` returns boxed `null`**
`app/src/main/java/io/apitomy/axiom/app/rest/AssistantResourceImpl.java` — `setAllowAll()`

```java
session.setAllowAll(data.getEnabled());
```

The OpenAPI spec marks `enabled` as required, but if the generated `getEnabled()` returns a boxed `Boolean` (which is common for OpenAPI-generated Java beans), a malformed request body missing the field could cause an unboxing NPE. This depends on whether the framework validates the `required` constraint before reaching this code. If there's no `@Valid` or `@NotNull` annotation on the parameter, a defensive null-check or explicit validation would be safer.

---

#### 💡 Suggestion

**3. Optimistic UI state may briefly diverge from server on SSE echo**
`ui/src/pages/AssistantSessionPage.tsx`

When toggling allow-all, the page both:
1. Calls the API (which emits an `allow_all_changed` SSE event)
2. Sets local state optimistically (`setAllowAll(true/false)`)

The SSE event later triggers `handleAllowAllChanged`, setting the same state again. This is harmless (idempotent) and is a valid optimistic-update pattern, but it's worth noting that if the API call succeeds but the SSE event arrives before the `setAllowAll` in the `then` block, there's no issue — both paths converge on the same value. No change needed, just calling out the intentional design.

---

#### 🔍 Nitpick

**4. Asymmetry between allow-all and rule-based auto-approval event visibility**
`app/src/main/java/io/apitomy/axiom/app/assistant/AssistantSession.java` — `handleAutoApproval()`

The allow-all path calls `addEvent(event)` before `respondToPermission()` so the permission request is visible in the UI stream. The existing rule-based path does **not** call `addEvent(event)` — only the `permission_resolved` event (from inside `respondToPermission`) reaches the stream. This is clearly intentional per the comment, but the behavioral difference could be confusing to future maintainers. A brief comment on the rule-based path noting this distinction (e.g., "// Rule-based approvals are silent — no permission_request event emitted") might help.

---

### Overall Assessment

**Looks good to merge.** The implementation is well thought out — `volatile` is correctly used for cross-thread visibility of the `allowAll` flag, the SSE event broadcast enables proper cross-tab sync, the confirmation modal provides appropriate guardrails for a security-sensitive toggle, and the visual indicators (warning color, border) make the active state clearly visible. The code follows established patterns in the codebase consistently.

The only actionable item is the missing user-visible error feedback (Warning #1), which is a minor UX gap. The potential NPE (Warning #2) depends on the code generation setup and may already be handled by framework validation.

PR #133, opened 2026-07-28
fix: defer navigation after session creation (#126)

### LINKED ISSUE #126: AI Assistant: Navigation to new session fails after template selection
## Bug Description

When creating a new AI Assistant session via **New Session → Template selection → Create Session**, the session is created successfully on the backend but the browser does not auto-navigate to `/assistant/{sessionId}`. The user remains on the sessions list page.

## Steps to Reproduce

1. Navigate to the AI Assistant page (`/assistant`)
2. Click **New Session**
3. Select a template from the template picker modal
4. Enter a name (or accept the default) and click **Create Session**
5. **Expected:** Browser navigates to the new session page at `/assistant/{id}`
6. **Actual:** Session is created, but the user stays on the sessions list page

## Root Cause Analysis

A timing conflict between React 19's concurrent rendering and React Router v7's `startTransition` wrapper.

In `CreateSessionModal.tsx`, `handleCreate()` fires multiple **urgent** state updates (closing modals, resetting form state) in the same synchronous block as the `navigate()` call (via `onSessionCreated`). React Router v7's `BrowserRouter` wraps all navigation state updates in `React.startTransition()` by default, making navigation a **low-priority transition**. React 19 processes the urgent modal-cleanup updates first, which triggers DOM mutations (PatternFly modal portal removal, backdrop cleanup, focus restoration) that can interrupt or supersede the deferred navigation transition.

### Contributing Factor

`AssistantSessionPage.tsx` has a catch block that navigates *back* to `/assistant` if the session fetch fails. Even if navigation partially succeeds, a race condition on the fetch could bounce the user back.

## Key Files

- `ui/src/main.tsx` — `<BrowserRouter>` without `useTransitions={false}`
- `ui/src/pages/AssistantPage.tsx` — `onSessionCreated` callback mixing `setState` + `navigate`
- `ui/src/components/assistant/CreateSessionModal.tsx` — `handleCreate()` with multiple urgent state updates before navigation
- `ui/src/pages/AssistantSessionPage.tsx` — bounce-back to `/assistant` on fetch failure

## Possible Fixes

**Option A (targeted):** Defer navigation into a `useEffect` in `AssistantPage.tsx` so it fires in a fresh render cycle after modal cleanup is complete:

```tsx
const [pendingSessionId, setPendingSessionId] = useState<string | null>(null);

useEffect(() => {
    if (pendingSessionId) {
        navigate(`/assistant/${pendingSessionId}`);
    }
}, [pendingSessionId, navigate]);

// In the callback:
onSessionCreated={(session) => {
    setIsTemplatePickerOpen(false);
    setPendingSessionId(session.id);
}}
```

**Option B (broader):** Disable `startTransition` wrapping for all navigation by adding `useTransitions={false}` to `<BrowserRouter>` in `main.tsx`.

**Option C (defense-in-depth):** In `CreateSessionModal.tsx`, call `onSessionCreated()` *without* first resetting modal state — the component is about to unmount via navigation anyway.

EricWittmann:
## Summary

Fixes #126 — after creating a new AI Assistant session, the browser failed to navigate to the session page due to a timing conflict between React 19's concurrent rendering and React Router v7's `startTransition` wrapper.

**Root cause:** `CreateSessionModal.handleCreate()` fired multiple urgent state updates (closing modals, resetting form state) in the same synchronous block as the `navigate()` call. React Router v7 wraps navigation in `startTransition`, making it low-priority. React 19 processed the urgent modal-cleanup updates first, triggering PatternFly modal DOM mutations that interrupted the deferred navigation.

**Changes:**

- **`AssistantPage.tsx`** — Defer navigation into a `useEffect` via a `pendingSessionId` state, so it fires in a fresh render cycle after modal cleanup is complete
- **`CreateSessionModal.tsx`** — Remove unnecessary modal state cleanup (`setIsNameModalOpen`, `setNewName`, `setSelectedTemplate`) before calling `onSessionCreated()` — the component unmounts on navigation anyway, and these urgent updates were competing with the deferred navigation transition

## Test plan

- [ ] Navigate to `/assistant`, click **New Session**, pick a template, enter a name, click **Create Session** — confirm browser navigates to `/assistant/{id}`
- [ ] Repeat several times to confirm reliability (original bug was a race condition)
- [ ] Test cancel flow: open template picker, cancel — confirm no navigation occurs
- [ ] Test navigating back to session list after viewing a session

### DIFF
--- a/ui/src/components/assistant/CreateSessionModal.tsx
+++ b/ui/src/components/assistant/CreateSessionModal.tsx
@@ -92,9 +92,6 @@ export function CreateSessionModal({
         try {
             const session = await createAssistantSession(
                 selectedTemplate.templateId, newName || undefined, projectId);
-            setIsNameModalOpen(false);
-            setNewName("");
-            setSelectedTemplate(null);
             onSessionCreated(session);
         } catch (err: unknown) {
             const e = err as { message?: string };
--- a/ui/src/pages/AssistantPage.tsx
+++ b/ui/src/pages/AssistantPage.tsx
@@ -52,6 +52,7 @@ export function AssistantPage() {
     const [sessions, setSessions] = useState<AssistantSessionInfo[]>([]);
     const [loading, setLoading] = useState(true);
     const [isTemplatePickerOpen, setIsTemplatePickerOpen] = useState(false);
+    const [pendingSessionId, setPendingSessionId] = useState<string | null>(null);
 
     const [filterName, setFilterName] = useState("");
     const [filterTemplateId, setFilterTemplateId] = useState("");
@@ -73,6 +74,12 @@ export function AssistantPage() {
     useEffect(() => {
         load();
     }, [load]);
+
+    useEffect(() => {
+        if (pendingSessionId) {
+            navigate(`/assistant/${pendingSessionId}`);
+        }
+    }, [pendingSessionId, navigate]);
 
     const filteredSessions = sessions.filter((s) => {
         if (filterName && !s.name.toLowerCase().includes(filterName.toLowerCase())) {
@@ -265,7 +272,7 @@ export function AssistantPage() {
                 onClose={() => setIsTemplatePickerOpen(false)}
                 onSessionCreated={(session) => {
                     setIsTemplatePickerOpen(false);
-                    navigate(`/assistant/${session.id}`);
+                    setPendingSessionId(session.id);
                 }}
             />
 


EricWittmann (commented):
## Code Review

This PR fixes a race condition where React 19's concurrent rendering would process urgent modal-cleanup state updates before a `startTransition`-wrapped `navigate()` call, causing navigation to be dropped after session creation. The fix defers navigation into a `useEffect` via a `pendingSessionId` state variable.

### Findings

#### ⚠️ Warning — Name modal's `isOpen` is decoupled from the parent's `isOpen` prop

**File:** `CreateSessionModal.tsx` (line 156)

The name modal's visibility is controlled solely by `isNameModalOpen`:
```tsx
<Modal isOpen={isNameModalOpen} ...>
```

With the removed cleanup lines (`setIsNameModalOpen(false)`, `setNewName("")`, `setSelectedTemplate(null)`), after a successful session creation:

1. Parent sets `isTemplatePickerOpen = false` and `pendingSessionId = session.id`
2. Re-render: the template picker modal hides (`isOpen && !isNameModalOpen` → `false`), but the name modal stays open (`isNameModalOpen` is still `true`)
3. Next effect cycle: `navigate()` fires and the component unmounts

In the **happy path**, this is a single imperceptible render frame — the navigation fires immediately after and the page unmounts. However, if navigation were to fail for any reason (e.g., a guard, an error), the name modal would remain orphaned on screen with no parent context controlling it.

The user can still dismiss it via the Cancel button (which calls `handleClose`, resetting all state), so this isn't a blocker, but it's worth being aware of. A more defensive approach would be to gate the name modal on the parent's `isOpen` prop as well:

```tsx
<Modal isOpen={isOpen && isNameModalOpen} ...>
```

This would make the cleanup removal completely safe regardless of whether navigation succeeds. Though this would be a slightly broader change.

#### 💡 Suggestion — Consider clearing `pendingSessionId` after navigation

**File:** `AssistantPage.tsx` (lines 78–82)

```tsx
useEffect(() => {
    if (pendingSessionId) {
        navigate(`/assistant/${pendingSessionId}`);
    }
}, [pendingSessionId, navigate]);
```

The `pendingSessionId` state is never explicitly cleared after navigation. In practice this is fine because navigation unmounts the component. However, for defensive coding and clarity of intent, resetting it after navigation makes the lifecycle more explicit:

```tsx
useEffect(() => {
    if (pendingSessionId) {
        navigate(`/assistant/${pendingSessionId}`);
        setPendingSessionId(null);
    }
}, [pendingSessionId, navigate]);
```

This would also protect against edge cases if React Router's behavior changes or if the component were ever kept mounted (e.g., via a layout route or `<Outlet>`).

### Overall Assessment

**Looks good to merge** ✅ with minor suggestions.

The fix is well-reasoned, minimal, and correctly addresses the root cause. The PR description is excellent — it clearly explains the React 19 / React Router v7 interaction causing the bug. The `useEffect`-based deferred navigation is a sound pattern for avoiding `startTransition` priority conflicts. The two suggestions above are defensive hardening rather than correctness issues for the current code.

PR #135, opened 2026-07-29
Add labels to Event Sources and propagate to auto-created Projects

### LINKED ISSUE #118: Add labels to Event Sources and propagate them to auto-created Projects
## Summary

Add support for labels on Event Sources so that when an event from that source triggers automatic
project creation, the event source's labels are propagated to the new project. This eliminates the
need to manually label projects after they are auto-created.

## Motivation

Currently, projects auto-created by the `PipelineOrchestrator` (via `createProjectFromEvent()`)
receive no labels. The only way to categorize them is through a manual `PUT /projects/{id}` update
after the fact. Since Event Sources represent a stable, configured identity (e.g. a specific GitHub
repo), they are a natural place to define labels that should automatically apply to all projects
originating from that source.

### 1. Add labels to Event Source

- Add a `labels` field (`List<String>`) to `EventSourceEntity` using `@ElementCollection` with a
  new `event_source_label` join table (following the existing pattern from `project_label`,
  `tool_label`, `report_label`).
- Add a new Flyway migration to create the `event_source_label` table.
- Update the OpenAPI spec to add `labels` to the `EventSource`, `NewEventSource`, and
  `UpdateEventSource` schemas.
- Update `EventSourcesResourceImpl` to handle labels on create and update.

### 2. Propagate labels to auto-created Projects

- In `PipelineOrchestrator.createProjectFromEvent()`, look up the `EventSourceEntity` using the
  event's `eventSourceId` and copy its labels onto the newly created `ProjectEntity`.
- Labels should only be set during initial project creation — subsequent events from the same
  source should not overwrite labels that may have been manually modified.

### 3. UI updates

- Add a labels field to the Event Source create/edit forms.
- Display labels on Event Source list/detail views (consistent with how project labels are shown).

## Design Notes

- Labels do **not** need to be stored on `EventEntity` — the orchestrator can resolve them from the
  event source at project-creation time, keeping the event model lean.
- This follows the existing label pattern: free-form `List<String>` stored via `@ElementCollection`.
- Events fired manually (via `POST /events`) may not have an `eventSourceId`, so label propagation
  should be conditional.

EricWittmann:
## Summary

- Add `labels` field to `EventSourceEntity` using `@ElementCollection` with a new
  `event_source_label` join table (Flyway V32), following the existing pattern from
  `project_label`, `tool_label`, and `report_label`.
- Update the OpenAPI spec to add `labels` to both `EventSource` and `NewEventSource` schemas.
- Handle labels in `EventSourcesResourceImpl` on create and update.
- Propagate event source labels to newly created projects in
  `PipelineOrchestrator.createProjectFromEvent()`. Labels are only set on initial
  project creation — subsequent events do not overwrite manually modified labels.
- Add labels column to the Event Sources list page using `ColoredLabel`.
- Add `LabelInput` to the create modal form.
- Add `LabelDisplay` + `EditLabelsModal` to the Event Source detail page.

Fixes #118

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -7,6 +7,7 @@ import io.apitomy.axiom.core.tracing.TraceService;
 import io.apitomy.axiom.core.entities.ActivityLogEntity;
 import io.apitomy.axiom.core.entities.EventEntity;
 import io.apitomy.axiom.core.entities.EventQueueEntity;
+import io.apitomy.axiom.core.entities.EventSourceEntity;
 import io.apitomy.axiom.core.entities.ProjectEntity;
 import io.apitomy.axiom.core.entities.TaskEntity;
 import io.apitomy.axiom.core.entities.ThreadEntryEntity;
@@ -543,6 +544,12 @@ public class PipelineOrchestrator {
         project.repository = event.repository != null ? event.repository : "unknown";
         project.createdOn = Instant.now();
         project.updatedOn = Instant.now();
+        if (event.eventSourceId != null) {
+            EventSourceEntity eventSource = EventSourceEntity.findById(event.eventSourceId);
+            if (eventSource != null && eventSource.labels != null && !eventSource.labels.isEmpty()) {
+                project.labels.addAll(eventSource.labels);
+            }
+        }
         project.persist();
 
         LOG.infof("Auto-created project %d for issue %s", project.id, event.issueRef);
--- a/app/src/main/java/io/apitomy/axiom/app/rest/EventSourcesResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/EventSourcesResourceImpl.java
@@ -107,6 +107,10 @@ public class EventSourcesResourceImpl implements EventResource {
         } else {
             entity.configuration = "{}";
         }
+        entity.labels.clear();
+        if (data.getLabels() != null) {
+            entity.labels.addAll(data.getLabels());
+        }
     }
 
     /**
@@ -147,6 +151,7 @@ public class EventSourcesResourceImpl implements EventResource {
                 // ignore parse errors
             }
         }
+        bean.setLabels(entity.labels);
         return bean;
     }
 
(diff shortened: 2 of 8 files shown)


EricWittmann (commented):
## Code Review — Add labels to Event Sources and propagate to auto-created Projects

This PR adds a `labels` field to `EventSourceEntity` (with a new Flyway migration), updates the OpenAPI spec and REST layer, propagates event source labels to auto-created projects in the pipeline orchestrator, and adds full UI support (list page labels column, create modal label input, detail page label display/edit).

### Findings

No critical or warning-level issues found. The implementation is clean and well-structured.

#### 💡 Suggestions

1. **`PipelineOrchestrator.createProjectFromEvent()` — extra DB query inside a transaction**
   (`app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java`, lines 547–551)

   The `EventSourceEntity.findById(event.eventSourceId)` call adds an extra database round-trip inside the `QuarkusTransaction.requiringNew()` block in `persistResults()`. Since `EventSourceEntity` uses `FetchType.EAGER` for labels, this will also join the `event_source_label` table. This is fine for correctness and the query will be fast (PK lookup), but worth noting — if this ever needs optimization, the event source labels could be pre-fetched or cached.

2. **`EventSourceDetailPage.tsx` — labels edited via `EditLabelsModal` don't mark the form as dirty**
   (`ui/src/pages/EventSourceDetailPage.tsx`, lines 182–189)

   The `EditLabelsModal` saves labels directly via `updateEventSource()` (bypassing the form's `dirty` state), while other fields (name, pollInterval, etc.) go through the `form` + `dirty` + Save Changes button flow. This is a deliberate design choice (matching the pattern used elsewhere with `EditLabelsModal`), but it does create two save paths. Not a bug — just noting the asymmetry. Users might be surprised that labels save immediately while other field changes require clicking "Save Changes."

#### 🔍 Nitpicks

1. **Migration file — index column order**
   (`app/src/main/resources/db/migration/V32__add_event_source_labels.sql`, line 7)

   The index is `(label, event_source_id)` which is optimal for filtering by label across event sources. The UNIQUE constraint `(event_source_id, label)` covers lookups by event source. This is a sensible setup — just confirming it's intentional and matches the query patterns.

### Overall Assessment

**Looks good to merge.** ✅

The changes follow the established patterns exactly (matching `ProjectEntity`, `ToolDefinitionEntity`, and `ReportEntity` for `@ElementCollection` labels). The Flyway migration is correct with proper FK/CASCADE, the OpenAPI spec is updated for both `EventSource` and `NewEventSource` schemas, the REST layer handles create/update/read correctly, the pipeline orchestrator correctly copies labels only on initial project creation, and the UI reuses existing `ColoredLabel`, `LabelInput`, `LabelDisplay`, and `EditLabelsModal` components consistently.

Well-scoped, consistent PR with no issues.

PR #136, opened 2026-07-29
Event Source label UI improvements and labels on Events page

EricWittmann:
## Summary

- **Event Source detail page**: Move Labels field below Name, and change the Edit Labels
  modal to update local form state (dirty flag) instead of saving to the server immediately.
  Labels are now persisted only when the user clicks "Save Changes", consistent with all
  other fields.
- **Event Source create modal**: Move Labels field to appear right after Name instead of at
  the bottom.
- **Events log page**: Add a Labels column showing gold-colored label chips inherited from
  the Event Source. Backend batch-loads Event Source labels to avoid N+1 queries.
- **OpenAPI spec**: Add `labels` array property to the `Event` schema.

## Test plan

- [ ] Edit labels on an Event Source detail page — verify changes are not saved until
      clicking "Save Changes"
- [ ] Create a new Event Source — verify Labels field appears after Name
- [ ] View the Events log page — verify labels column shows chips from the Event Source
- [ ] Verify events without an Event Source show "—" in the Labels column

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/rest/EventsResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/EventsResourceImpl.java
@@ -7,6 +7,7 @@ import io.apitomy.axiom.api.beans.NewEvent;
 import io.apitomy.axiom.api.beans.Trace;
 import io.apitomy.axiom.core.entities.TraceEntity;
 import io.apitomy.axiom.core.entities.EventEntity;
+import io.apitomy.axiom.core.entities.EventSourceEntity;
 import io.apitomy.axiom.events.core.EventService;
 import io.quarkus.panache.common.Page;
 import io.quarkus.panache.common.Sort;
@@ -16,10 +17,13 @@ import jakarta.inject.Inject;
 import jakarta.ws.rs.WebApplicationException;
 
 import java.math.BigInteger;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Implementation of the Events REST API. Provides paginated, filterable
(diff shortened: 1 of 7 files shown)


EricWittmann (commented):
## Code Review Summary

This PR makes three related improvements: (1) labels on the Event Source detail page now update local form state and are saved with the "Save Changes" button instead of immediately, (2) the Labels field is repositioned below Name in both the detail and create forms, and (3) a new Labels column is added to the Events log page, backed by batch-loaded Event Source labels on the backend.

---

### Findings

#### 💡 Suggestion

**Duplicated `loadEventSourceLabels` method across two resource classes**
- `EventsResourceImpl.java` (lines 127-138) and `ProjectsResourceImpl.java` (lines 553-564)

The batch-loading logic is identical in both classes. Consider extracting it into a shared helper (e.g., a `@ApplicationScoped` service or a static utility method) to keep the logic in one place. This avoids future drift if the loading logic needs to change (e.g., adding caching or handling soft-deleted sources).

#### 🔍 Nitpick

**Potential `NullPointerException` in `Collectors.toMap` if `es.labels` is ever null**
- `EventsResourceImpl.java` line 137: `.collect(Collectors.toMap(es -> es.id, es -> es.labels))`
- `ProjectsResourceImpl.java` line 563: same pattern

`EventSourceEntity.labels` is initialized to `new ArrayList<>()` and uses `FetchType.EAGER`, so under normal Hibernate behavior this will never be null. However, a defensive `es -> es.labels != null ? es.labels : Collections.emptyList()` would guard against edge cases with unusual deserialization or direct DB manipulation. Very low risk — noting for completeness.

#### 🔍 Nitpick

**React key on label chips uses the label string value**
- `EventsPage.tsx` line 190: `<Label key={lbl} ...>`

If an event source somehow had duplicate labels, this would cause a React key collision warning. Since labels are typically unique within a source, this is unlikely to be hit, but `key={idx}` or a compound key like `key={\`${event.id}-${lbl}\`}` would be more robust.

---

### Overall Assessment

✅ **Looks good to merge.** This is a clean, well-structured PR. Highlights:

- **Batch loading** via `loadEventSourceLabels` correctly avoids N+1 queries for the list endpoints; the single-entity `lookupEventSourceLabels` is appropriate for `getEvent`/`fireEvent`.
- **Form state management** on the detail page is now consistent — labels flow through `updateForm` → dirty flag → save, matching all other fields.
- **OpenAPI spec** is updated to include the new `labels` array on the `Event` schema.
- The UI field reordering (Labels below Name) is applied consistently in both the detail and create views.

The duplicated helper method is the only real suggestion; everything else is minor. Nice work!

Issue #142, opened 2026-07-31
Duplicate Max Steps and Max Budget fields on Action Type Info tab


EricWittmann:
## Bug

The **Max Steps** and **Max Budget (USD)** form fields are rendered twice on the Action Type detail page (Info tab) when execution mode is set to "actor".

## Root Cause

In `ui/src/pages/ActionTypeDetailPage.tsx`, the `InfoTab` component contains two identical blocks of form fields for Max Steps and Max Budget. One block appears before the Model field and a duplicate block appears after it.

## Expected Behavior

Each field should appear exactly once.

## Fix

Remove the duplicate block (the second occurrence, after the Model field).

EricWittmann:
Fixed in:
* https://github.com/Apitomy/apitomy-axiom/pull/143

PR #143, opened 2026-07-31
Remove duplicate Max Steps and Max Budget fields

EricWittmann:
## Summary\n\nFixes #142 — the Max Steps and Max Budget (USD) form fields were rendered twice on the Action Type detail page (Info tab) when execution mode was set to "actor". Removed the duplicate block that appeared after the Model field.\n\n## Test Plan\n\n- Open the Action Type detail page with execution mode set to "actor"\n- Confirm Max Steps and Max Budget (USD) each appear exactly once\n- Confirm the fields still accept input and clear to global default

### DIFF
--- a/ui/src/pages/ActionTypeDetailPage.tsx
+++ b/ui/src/pages/ActionTypeDetailPage.tsx
@@ -467,34 +467,6 @@ function InfoTab({ form, updateForm, availableModels, availableEngines }: {
                     </FormSelect>
                 </FormGroup>
             )}
-            {form.executionMode === "actor" && (
-                <FormGroup label="Max Steps" fieldId="maxSteps">
-                    <HelperText>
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Review Summary

This PR removes a duplicate rendering of the **Max Steps** and **Max Budget (USD)** form fields from the `InfoTab` component in `ActionTypeDetailPage.tsx`. The same two fields were rendered once after the AI Engine selector (lines 400–427, kept) and again after the Model selector (lines 467–497 in the old code, removed). The deletion is correct and straightforward.

## Findings

No issues found. The removed block was a character-for-character duplicate of the retained block — same field IDs, same helper text, same `onChange` handlers. Removing it is the right fix.

## Overall Assessment

✅ **Looks good to merge.** Clean, minimal, single-file change that fixes a clear UI bug with no risk of side effects.

PR #163, opened 2026-08-06
Add Disk Usage Breakdown dashboard widget

### LINKED ISSUE #155: New dashboard widget: Disk Usage Breakdown
## Summary

Add a new dashboard widget that shows disk usage breakdown. This was identified in the
custom dashboards design spec as a future extension widget.

## Details

The widget should display storage consumption across the system, broken down by
category (e.g. repositories, reports, logs, attachments). A bar chart or donut chart
visualization would be appropriate.

## Widget Registry Entry

- **Type:** `disk-usage`
- **Category:** System
- **Default size:** TBD
- **Config options:** TBD

EricWittmann:
## Summary

Adds a new **Disk Usage Breakdown** dashboard widget that displays storage consumption broken down by project using a bar chart visualization.

Fixes #155

## Changes

- **New file:** `ui/src/components/dashboards/widgets/DiskUsageWidget.tsx` — The widget component that fetches disk usage data via the existing `fetchDiskUsage` API and renders it as a bar chart using PatternFly's Victory chart components. Includes a total disk usage summary at the top and hover tooltips showing formatted byte values.
- **Modified file:** `ui/src/components/dashboards/widgets/index.ts` — Added the import for the new widget so it registers itself in the widget registry on startup.

## Widget Details

- **Type:** `disk-usage`
- **Category:** System
- **Default size:** 4×3
- **Min size:** 3×3
- **Config options:** `maxProjects` (number, 1–20, default 10)

## Implementation Notes

- Follows the same pattern as the existing `AiCostByProjectWidget` (bar chart, cancellation-safe effect, `WidgetError` fallback, empty-state message).
- Uses the existing `fetchDiskUsage` API and `formatBytes` utility from `config/api.ts` — no backend changes needed.
- Projects with zero disk usage are filtered out and the remaining are sorted by usage descending.

### DIFF
--- a/ui/src/components/dashboards/widgets/DiskUsageWidget.tsx
+++ b/ui/src/components/dashboards/widgets/DiskUsageWidget.tsx
@@ -1,0 +1,84 @@
+import { useState, useEffect } from "react";
+import { ChartBar, ChartGroup, ChartVoronoiContainer, Chart, ChartAxis } from "@patternfly/react-charts/victory";
+import { fetchDiskUsage, formatBytes } from "../../../config/api";
+import { registerWidget, type WidgetProps } from "../widget-registry";
+import { WidgetError } from "../WidgetError";
+
+interface ProjectDisk {
+    name: string;
+    bytes: number;
+}
+
+function DiskUsageWidget({ config }: WidgetProps) {
+    const [data, setData] = useState<ProjectDisk[]>([]);
+    const [totalBytes, setTotalBytes] = useState(0);
+    const [error, setError] = useState(false);
+
+    const maxProjects = Number(config.maxProjects) || 10;
+
+    useEffect(() => {
+        let cancelled = false;
+        fetchDiskUsage(1, maxProjects)
+            .then(result => {
+                if (cancelled) return;
+                const items = result.items
+                    .map(p => ({ name: p.projectName, bytes: p.diskUsageBytes }))
+                    .filter(p => p.bytes > 0)
+                    .sort((a, b) => b.bytes - a.bytes);
+                setData(items);
+                setTotalBytes(result.totalDiskUsageBytes);
+            })
+            .catch(() => { if (!cancelled) setError(true); });
+        return () => { cancelled = true; };
+    }, [maxProjects]);
+
+    if (error) return <WidgetError />;
+
+    if (data.length === 0) {
+        return <p style={{ padding: "16px", textAlign: "center" }}>No disk usage data.</p>;
+    }
+
+    const chartData = data.map(d => ({
+        x: d.name,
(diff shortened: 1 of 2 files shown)


EricWittmann (commented):
## Code Review — Disk Usage Breakdown Widget

Clean, well-structured PR that adds a new `disk-usage` dashboard widget following the established bar-chart widget pattern (`AiCostByProjectWidget`). The implementation reuses existing API functions and utilities, and the registration/index wiring is correct.

---

### 🔍 Findings

#### Suggestion

**1. Empty state shown during initial loading** — `DiskUsageWidget.tsx` L37-38

`data` is initialised as `[]`, so the component renders "No disk usage data." on first mount before the `fetchDiskUsage` promise resolves. This causes a brief flash of the empty state. The same pattern exists in `AiCostByProjectWidget`, so this is **consistent with the codebase** — but adding a `loading` boolean (default `true`, set to `false` after the fetch settles) and rendering a spinner/skeleton during that phase would be a nice UX improvement across all chart widgets. Non-blocking.

**2. Top-N correctness depends on server-side ordering** — `DiskUsageWidget.tsx` L21, L27

```ts
fetchDiskUsage(1, maxProjects)   // fetch page 1 only
  .sort((a, b) => b.bytes - a.bytes);  // client-side sort
```

The client-side sort only reorders the items returned on page 1. If the backend doesn't guarantee results are ordered by `diskUsageBytes` descending, the chart might miss higher-usage projects that fall on later pages. If the `/usage/disk` endpoint already sorts by usage (likely, given its purpose), this is fine — but it might be worth confirming. An alternative would be requesting a larger page and slicing client-side, similar to how `AiCostByProjectWidget` fetches up to 50 projects then slices to `maxProjects`.

#### Nitpick

**3. "Total" label could be more explicit** — `DiskUsageWidget.tsx` L52

`result.totalDiskUsageBytes` is an aggregate across **all** projects, but the bar chart only displays up to `maxProjects`. When a user sees e.g. 5 bars summing to 500 MB but the "Total" reads 2 GB, it may be momentarily confusing. Consider labeling it **"Total (all projects)"** to set expectations.

**4. `labels` prop unused** — `DiskUsageWidget.tsx` L12

The `WidgetProps` interface includes `labels: string[]`, and the comparable `AiCostByProjectWidget` uses it for filtering. Since `fetchDiskUsage` doesn't accept a labels parameter, this is expected — just noting it as a potential future enhancement if the backend adds label-based filtering to the disk usage endpoint.

---

### ✅ Overall

This is a **clean PR, good to merge**. The implementation is correct, consistent with existing widget patterns, and properly wired into the widget registry. The suggestions above are all non-blocking improvements that could be tackled separately.

PR #164, opened 2026-08-06
Fix AI Cost Summary widget "Error loading data"

EricWittmann:
## Summary
- The AI Cost Summary and AI Cost by Project dashboard widgets send `filterDateFrom` as a full
  ISO-8601 datetime string (e.g. `2026-07-30T12:34:56.789Z`) via `new Date(...).toISOString()`
- The backend `UsageResourceImpl` parsed this with `LocalDate.parse()`, which only accepts
  date-only format (`2026-07-30`), causing a `DateTimeParseException` and a 500 error
- The standalone AI Usage page was unaffected because its date inputs produce date-only strings
- Added a `parseInstant()` helper that detects the format and parses accordingly

## Test plan
- [ ] Verify the AI Cost Summary widget loads data on the dashboard
- [ ] Verify the AI Cost by Project widget loads data on the dashboard
- [ ] Verify the AI Usage page date filters still work correctly
- [ ] Test with different time windows (24h, 7d, 30d, 90d) on both widgets

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
@@ -69,14 +69,15 @@ public class UsageResourceImpl implements UsageResource {
             params.put("actionType", "%" + filterActionType.toLowerCase() + "%");
         }
         if (filterDateFrom != null && !filterDateFrom.isBlank()) {
-            Instant fromInstant = LocalDate.parse(filterDateFrom)
-                    .atStartOfDay(ZoneId.systemDefault()).toInstant();
+            Instant fromInstant = parseInstant(filterDateFrom);
             hql.append(" and createdOn >= :dateFrom");
             params.put("dateFrom", fromInstant);
         }
         if (filterDateTo != null && !filterDateTo.isBlank()) {
-            Instant toInstant = LocalDate.parse(filterDateTo)
-                    .plusDays(1).atStartOfDay(ZoneId.systemDefault()).toInstant();
+            Instant toInstant = filterDateTo.contains("T")
+                    ? Instant.parse(filterDateTo)
+                    : LocalDate.parse(filterDateTo).plusDays(1)
+                            .atStartOfDay(ZoneId.systemDefault()).toInstant();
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review

This PR fixes a 500 error caused by the AI Cost Summary/Project dashboard widgets sending full ISO-8601 datetime strings (e.g. `2026-07-30T12:34:56.789Z`) to an endpoint that only accepted date-only strings (`2026-07-30`). A `parseInstant()` helper is added to handle both formats.

### Findings

#### 💡 Suggestion — `filterDateTo` duplicates logic instead of reusing `parseInstant()`

**File:** `UsageResourceImpl.java`, lines 77-80

The `filterDateFrom` block uses the new `parseInstant()` helper, but the `filterDateTo` block uses inline logic with the same `contains("T")` / `Instant.parse()` / `LocalDate.parse()` pattern. I understand the difference — `filterDateTo` needs `plusDays(1)` for date-only strings to make the end date inclusive — but this could be tidied up. For example, a small refactor like:

```java
Instant toInstant = parseInstant(filterDateTo);
if (!filterDateTo.contains("T")) {
    // For date-only input, advance by one day to make the end date inclusive
    toInstant = toInstant.plus(1, ChronoUnit.DAYS);
}
```

Or alternatively, an overloaded helper / a parameter for the `plusDays` adjustment. This keeps the format-detection logic in a single place.

#### 🔍 Nitpick — Consider defensive error handling for malformed date inputs

Both `Instant.parse()` and `LocalDate.parse()` will throw `DateTimeParseException` on malformed input, resulting in an unhandled 500 error. This is pre-existing behavior and not introduced by this PR, but since you're already touching this code, wrapping the parse in a try-catch with a meaningful 400-level error response would be a nice improvement. Not blocking.

### Overall Assessment

**Looks good to merge.** ✅

This is a clean, well-scoped bug fix that correctly addresses the root cause. The `contains("T")` heuristic is a reliable way to distinguish the two date formats in this context, and the semantic difference between `filterDateFrom` (exact instant) and `filterDateTo` (inclusive end-of-day for date-only) is handled correctly. The suggestions above are minor improvements, not blockers.

PR #165, opened 2026-08-07
Add Disk Usage Breakdown dashboard widget

EricWittmann:
## Summary

Adds a new **Disk Usage Breakdown** dashboard widget that displays storage consumption broken down by project using a bar chart visualization.

Fixes #163

## Changes

- **New file:** `ui/src/components/dashboards/widgets/DiskUsageWidget.tsx` — A widget component that fetches disk usage data via the existing `fetchDiskUsage` API and renders it as a bar chart using PatternFly's Victory chart components. Includes a total disk usage summary at the top and hover tooltips showing formatted byte values.
- **Modified file:** `ui/src/components/dashboards/widgets/index.ts` — Added the import for the new widget so it registers itself in the widget registry on startup.

## Widget Details

- **Type:** `disk-usage`
- **Category:** System
- **Default size:** 4×3
- **Min size:** 3×3
- **Config options:** `maxProjects` (number, 1–20, default 10)

## Implementation Notes

- Follows the same pattern as the existing `AiCostByProjectWidget` (bar chart, cancellation-safe effect, `WidgetError` fallback, empty-state message).
- Uses the existing `fetchDiskUsage` API and `formatBytes` utility — no backend changes needed.
- Projects with zero disk usage are filtered out and the remaining are sorted by usage descending.

### DIFF
--- a/ui/src/components/dashboards/widgets/DiskUsageWidget.tsx
+++ b/ui/src/components/dashboards/widgets/DiskUsageWidget.tsx
@@ -1,0 +1,75 @@
+import { useState, useEffect } from "react";
+import { ChartBar, ChartGroup, ChartVoronoiContainer, Chart, ChartAxis } from "@patternfly/react-charts/victory";
+import { fetchDiskUsage, formatBytes } from "../../../config/api";
+import { registerWidget, type WidgetProps } from "../widget-registry";
+import { WidgetError } from "../WidgetError";
+
+interface ProjectDiskUsage {
+    name: string;
+    bytes: number;
+}
+
+function DiskUsageWidget({ config }: WidgetProps) {
+    const [data, setData] = useState<ProjectDiskUsage[]>([]);
+    const [totalBytes, setTotalBytes] = useState(0);
+    const [error, setError] = useState(false);
+    const maxProjects = Number(config.maxProjects) || 10;
+
+    useEffect(() => {
+        let cancelled = false;
+        fetchDiskUsage(1, 50)
+            .then((result) => {
+                if (cancelled) return;
+                const projects = result.items
+                    .filter(p => p.diskUsageBytes > 0)
+                    .sort((a, b) => b.diskUsageBytes - a.diskUsageBytes)
+                    .slice(0, maxProjects)
+                    .map(p => ({ name: p.projectName, bytes: p.diskUsageBytes }));
+                setData(projects);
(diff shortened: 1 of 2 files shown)


EricWittmann (commented):
## Code Review — Disk Usage Breakdown Widget

Adds a new `disk-usage` dashboard widget that fetches per-project storage data via `fetchDiskUsage` and renders a PatternFly bar chart with a total summary, closely following the existing `AiCostByProjectWidget` pattern.

---

### Findings

#### 💡 Suggestion — Hardcoded page size may miss high-usage projects

**File:** `DiskUsageWidget.tsx`, line 20

```ts
fetchDiskUsage(1, 50)
```

The widget fetches the first 50 items from the paginated disk usage API, then sorts client-side by `diskUsageBytes` descending. If there are more than 50 projects, the top consumers could fall outside this page and never appear in the chart. The `totalBytes` summary (sourced from `result.totalDiskUsageBytes`) would still be correct, but the chart could be misleading.

Since `maxProjects` caps at 20, the 50-item buffer is probably sufficient for most deployments. However, if the server could return results pre-sorted by `diskUsageBytes desc`, you could drop the client-side sort and be confident the top N are always present. Worth keeping in mind as the user base grows.

#### 💡 Suggestion — Initial empty state is indistinguishable from loading

**File:** `DiskUsageWidget.tsx`, lines 37–39

```tsx
if (data.length === 0) {
    return <p style={{ padding: "16px", textAlign: "center" }}>No disk usage data.</p>;
}
```

Since `data` starts as `[]`, users briefly see "No disk usage data." before the fetch completes. Adding a `loading` state (or a skeleton/spinner) would eliminate the flash. That said, `AiCostByProjectWidget` has the same behavior, so this is consistent with the existing codebase — just noting it as a potential UX improvement to address across all widgets in a future pass.

---

### Overall Assessment

**Looks good to merge.** ✅

This is a clean, well-scoped PR. The implementation closely follows the established widget conventions (`AiCostByProjectWidget`): cancellation-safe effect, `WidgetError` fallback, consistent chart setup, and proper widget registration. The API types (`DiskUsageProject`, `DiskUsageSearchResults`) and utilities (`formatBytes`) are reused correctly with no backend changes needed. The two suggestions above are minor and neither blocks merging.

EricWittmann:
Close as duplicate of:
* https://github.com/Apitomy/apitomy-axiom/pull/163

PR #166, opened 2026-08-07
feat: user-configurable event source filtering

### LINKED ISSUE #127: feat: user-configurable event source filtering
## Summary

Event sources are currently all-or-nothing: configure one, and every issue, comment, PR, and
review comment gets ingested and sent to the AI Manager for evaluation. On active repos this
means the Manager — an LLM call that costs real money and takes seconds — is invoked on a
large volume of noise the user never cared about.

PR #122 added hardcoded bot and slash-command pre-filters in `PipelineOrchestrator`, which
was a pragmatic fix for the immediate cost/rate-limit pain (see #121). But it's not
user-configurable, not extensible, and only covers one class of noise. This issue proposes a
general-purpose, per-event-source filtering layer that gives users control over which events
flow through to the Manager.

## Motivation

- **Cost** — Every event reaching the Manager costs money. On one repo, 50%+ of events were
  filterable noise (bots, CI chatter, slash commands).
- **Latency** — Events process sequentially from a queue. Noise events block the queue and
  delay events the user cares about.
- **User control** — Different teams care about different things. One team may only want
  `pr-created` and `pr-merged`; another wants all issue events but nothing from PRs.
  Hardcoding can't serve both.
- **Replaces hardcoded logic** — The `shouldSkipEvent()` rules from PR #122 would become the
  default filter configuration rather than baked-in logic.

### Filter model

Filters are **per event source**, stored in the existing `configuration` JSON blob (no schema
migration needed). A simple include/exclude model:

```json
{
  "owner": "Apitomy",
  "name": "apitomy-axiom",
  "filters": {
    "eventTypes": {
      "include": ["issue-created", "issue-updated", "comment-added", "pr-created"],
      "exclude": []
    },
    "actors": {
      "exclude": ["*[bot]"]
    }
  }
}
```

**Semantics:**
- An empty `include` list means "allow all" (then `exclude` removes from that set).
- A non-empty `include` list means "only allow these" (then `exclude` further narrows).
- Glob patterns supported for actors (`*[bot]`), branches (`release/*`), labels, etc.

### Where filtering happens

Filtering should happen **at the poller level** (before ingestion), so filtered events never
enter the DB queue or consume processing time. The current `shouldSkipEvent()` in
`PipelineOrchestrator` remains as a safety net but the bulk of filtering moves earlier.

To preserve observability, the poll log (which already tracks `eventsIngested`) could gain an
`eventsFiltered` count.

### Filter dimensions

Start with the two highest-value dimensions:
- **Event type** — e.g. only ingest `issue-created` and `pr-created`, skip everything else
- **Actor patterns** — e.g. exclude `*[bot]` (subsumes the hardcoded bot filter)

Future increments could add:
- **Branch patterns** — e.g. only PRs targeting `main` or `release/*`
- **Label patterns** — e.g. only issues with `bug` or `feature` labels
- **Content patterns** — e.g. skip comments matching slash-command patterns

### Filter debugging and observability

Users need a way to understand why events were allowed through or blocked by filters. Without
this, filter misconfiguration is invisible — events silently disappear or unexpectedly flood
the Manager, and users have no way to diagnose the problem.

**Requirements:**
- Each poll cycle should log a summary of filter activity: how many events were evaluated,
  how many passed, and how many were filtered out.
- Filtered events should be recorded with enough detail to understand **which filter rule**
  caused the block and **what the event was** (event type, actor, issue ref). This could be
  a structured detail field on the existing poll log, or a dedicated filter log.
- The UI should surface filter activity — either in the existing Poll Logs tab on the event
  source detail page, or as a new "Filter Log" tab — so users can see at a glance what's
  being blocked and why.
- Consider a "dry run" or "test mode" capability where users can preview what their filter
  configuration would do against recent events without actually changing the live behavior.

## Scope of Work

- **OpenAPI spec** — Add filter schema definitions, update `EventSource`/`NewEventSource`
- **Backend** — Shared filter evaluation utility, integrate into `GitHubPoller` and
  `JiraPoller`
- **Backend** — Filter debug logging: per-event filter decisions with matched rule detail
- **Frontend** — Filter configuration UI in event source create/edit forms
- **Frontend** — Filter activity visibility in event source detail page
- **Tests** — Unit tests for filter evaluation logic
- **Migration** — Update `shouldSkipEvent()` to defer to configured filters with sensible
  defaults

## Related

- #121 — The subprocess hang issue that prompted the bot pre-filter work
- #122 — PR that added hardcoded bot/slash-command pre-filtering (this feature supersedes
  that approach)

EricWittmann:
Fixes #127

## Summary

- **Configurable filters:** Per-event-source include/exclude filter rules stored as a new `filters` JSON column on `event_source`. Supports two rule types: `event-type` (glob against the event type string) and `payload` (JSON Pointer + glob against a payload field value).
- **Replaces hardcoded logic:** Removes `shouldSkipEvent()` and `SLASH_COMMANDS` from `PipelineOrchestrator`. The old bot and slash-command filters are migrated into default filter rules via DB migration V34, applied to all existing and new event sources.
- **Dry-run endpoint:** `POST /api/event-sources/filters/dry-run` fetches live events from the source API (GitHub/Jira), classifies them using extracted poller logic, and evaluates proposed filters — letting users preview what would be allowed or blocked before saving.
- **Frontend Filters tab:** New tab on the event source detail page with include/exclude rule tables, inline add/delete, save, and a "Test Filters" button that shows dry-run results with green/red allowed/blocked labels and summary counts.

## Changes

| Area | What changed |
|------|-------------|
| OpenAPI | Added `EventSourceFilterRule`, `EventSourceFilters`, `FilterDryRunRequest/Response/Result` schemas; `filters` on `EventSource`/`NewEventSource`; dry-run endpoint |
| Core | `EventFilterEvaluator` with wildcard matching (17 unit tests), filter model records |
| DB | V34 migration: `filters TEXT` column with default exclude rules |
| App | REST serialization, `PipelineOrchestrator` filter integration, dry-run endpoint implementation |
| Events | Extracted `GitHubEventClassifier`/`JiraEventClassifier` from pollers; `GitHubDryRunService`/`JiraDryRunService` |
| UI | TypeScript filter types, `dryRunFilters()` API call, `FiltersTab` component |

## Test plan

- [ ] Build passes (`build.sh`) — 282 tests, 0 failures
- [ ] UI builds (`npm run build`) — TypeScript + Vite clean
- [ ] Create new event source — verify default filters are populated
- [ ] Edit filters via Filters tab — verify save round-trips correctly
- [ ] Test Filters dry-run — verify live events are fetched and filter verdicts displayed
- [ ] Trigger event ingestion — verify filtered events show `event-pre-filtered` in activity log with matched rule detail
- [ ] Verify bot events blocked by default `*[bot]` rules
- [ ] Verify slash-command comments blocked by default `/*` rule

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
+++ b/app/src/main/java/io/apitomy/axiom/app/PipelineOrchestrator.java
@@ -2,6 +2,9 @@ package io.apitomy.axiom.app;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.apitomy.axiom.core.filters.EventFilterEvaluator;
+import io.apitomy.axiom.core.filters.EventSourceFilters;
+import io.apitomy.axiom.core.filters.FilterResult;
 import io.apitomy.axiom.core.tracing.TraceContext;
 import io.apitomy.axiom.core.tracing.TraceService;
 import io.apitomy.axiom.core.entities.ActivityLogEntity;
@@ -42,10 +45,8 @@ public class PipelineOrchestrator {
 
     private static final Logger LOG = Logger.getLogger(PipelineOrchestrator.class);
 
-    private static final Set<String> SLASH_COMMANDS = Set.of(
-            "/ready", "/retry", "/accept", "/reject", "/disable-tests",
-            "/enable-tests", "/unstale", "/skip-review", "/auto-merge", "/merge"
-    );
+    @Inject
+    EventFilterEvaluator filterEvaluator;
 
     @Inject
     ManagerService managerService;
@@ -104,43 +105,27 @@ public class PipelineOrchestrator {
     }
 
     /**
-     * Returns a skip reason if the event should be filtered out without
-     * invoking the AI Manager, or {@code null} if the event should be processed.
+     * Evaluates event against the filters configured on its source.
+     *
+     * @param event the event to evaluate
+     * @return the filter evaluation result
      */
-    private String shouldSkipEvent(EventEntity event) {
-        if (event.payload == null || event.payload.isBlank()) {
-            return null;
+    private FilterResult evaluateFilters(EventEntity event) {
+        if (event.eventSourceId == null) {
+            return FilterResult.ALLOWED;
+        }
+        EventSourceEntity source = EventSourceEntity.findById(event.eventSourceId);
+        if (source == null || source.filters == null) {
+            return FilterResult.ALLOWED;
         }
         try {
+            EventSourceFilters filters = objectMapper.readValue(source.filters, EventSourceFilters.class);
             JsonNode payload = objectMapper.readTree(event.payload);
-
-            // Extract the author login from the event payload
-            String login = null;
-            JsonNode comment = payload.path("comment");
-            if (!comment.isMissingNode()) {
-                login = comment.path("user").path("login").asText(null);
-            }
-            if (login == null) {
-                login = payload.path("user").path("login").asText(null);
-            }
-
-            // Skip events from GitHub App bot accounts (all use the [bot] suffix)
-            if (login != null && login.endsWith("[bot]")) {
-                return "bot activity from " + login;
-            }
-
-            // Skip slash command comments (lifecycle orchestrator handles these)
-            if ("comment-added".equals(event.eventType) && !comment.isMissingNode()) {
-                String body = comment.path("body").asText("").trim();
-                if (SLASH_COMMANDS.stream().anyMatch(cmd ->
-                        body.equals(cmd) || body.startsWith(cmd + " "))) {
-                    return "slash command: " + body.split("\\s+")[0];
-                }
-            }
+            return filterEvaluator.evaluate(filters, event.eventType, payload);
         } catch (Exception e) {
-            LOG.tracef("Failed to parse event payload for pre-filtering: %s", e.getMessage());
+            LOG.warnf("Failed to evaluate filters for event %d: %s", event.id, e.getMessage());
+            return FilterResult.ALLOWED;
         }
-        return null;
     }
 
     /**
@@ -170,13 +155,13 @@ public class PipelineOrchestrator {
         LOG.infof("Processing event %d: %s [%s] from %s",
                 event.id, event.eventType, event.issueRef, event.source);
 
-        // Pre-filter: skip events from known bots without invoking the AI Manager
-        String skipReason = shouldSkipEvent(event);
-        if (skipReason != null) {
-            LOG.infof("Pre-filtered event %d: %s", event.id, skipReason);
+        // Pre-filter: evaluate event against source filters before invoking the AI Manager
+        FilterResult filterResult = evaluateFilters(event);
+        if (!filterResult.allowed()) {
+            LOG.infof("Pre-filtered event %d: %s", event.id, filterResult.matchedRule());
             QuarkusTransaction.requiringNew().run(() -> {
                 logActivity(null, null, event.id, "event-pre-filtered",
-                        "Event pre-filtered: " + event.eventType + " — " + skipReason);
+                        "Event pre-filtered: " + event.eventType + " — " + filterResult.matchedRule());
                 markQueueEntry(queueEntryId, "completed");
             });
             return;
--- a/app/src/main/java/io/apitomy/axiom/app/rest/EventSourcesResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/EventSourcesResourceImpl.java
@@ -52,6 +52,13 @@ public class EventSourcesResourceImpl implements EventResource {
     public EventSource createEventSource(NewEventSource data) {
         EventSourceEntity entity = new EventSourceEntity();
         applyFields(entity, data);
+        if (entity.filters == null) {
+            try {
+                entity.filters = objectMapper.writeValueAsString(defaultFilters());
+            } catch (Exception e) {
+                entity.filters = null;
+            }
+        }
         entity.persist();
         return toBean(entity);
     }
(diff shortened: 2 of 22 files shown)


EricWittmann (commented):
## Code Review: User-configurable event source filtering

This PR adds a comprehensive event source filtering system, including an include/exclude filter rule model (with glob patterns and JSON Pointer payload matching), an `EventFilterEvaluator` in core, a new dry-run API endpoint (`POST /event-sources/filters/dry-run`), new `GitHubDryRunService` and `JiraDryRunService` for live-testing filters, a `Filters` tab in the Event Source detail page UI, and default exclude rules for bot activity and slash commands on new event sources.

---

### 🔴 Critical

**1. Filters are never applied during actual event polling**

The filter infrastructure (entity storage, evaluator, dry-run testing, UI) is fully built out, but the filters are **never checked during real event ingestion**. Both `GitHubPoller.pollRepository()` and `JiraPoller.pollProject()` call `eventService.ingestEvent()` without evaluating the event against the source's filter configuration. The `EventService.ingestEvent()` method also has no filter awareness.

This means users can configure and test filters, see them working in dry-run, save them — and then watch as all events are ingested unfiltered anyway.

The pollers need to:
1. Deserialize `source.filters` into the core `EventSourceFilters` model
2. Build the payload `JsonNode` before calling `ingestEvent()`
3. Call `filterEvaluator.evaluate()` and skip ingestion for blocked events
4. Log filtered-out events in the poll log for observability

**Files:** `events/github/.../GitHubPoller.java`, `events/jira/.../JiraPoller.java`, `events/core/.../EventService.java`

---

### ⚠️ Warning

**2. Dry-run endpoint fails for event sources using default authentication**

`resolveSecretValue()` in `EventSourcesResourceImpl` (lines 373–385) only looks up explicitly-named secrets from the database. Unlike the pollers' `resolveToken()` methods, it does **not** fall back to default provider secrets (`GH_TOKEN`, `GITHUB_TOKEN`, `JIRA_API_TOKEN`) or environment variables.

When a user creates an event source with the default "auto-detect" authentication (no `secretName`), dry-run will throw a 401 error even though actual polling works fine with the same configuration.

```java
// EventSourcesResourceImpl.resolveSecretValue() — only named lookup
if (secretName != null && !secretName.isBlank()) {
    SecretEntity secret = SecretEntity.find("name", secretName).firstResult();
    ...
}
return null;  // ← triggers 401 in fetchDryRunEvents()

// vs GitHubPoller.resolveToken() — has fallbacks
for (String name : List.of("GH_TOKEN", "GITHUB_TOKEN")) { ... }
String token = System.getenv("GH_TOKEN"); // etc.
```

**File:** `app/.../EventSourcesResourceImpl.java` lines 322–385

**3. Missing Flyway migration for `filters` column**

The `EventSourceEntity` adds a new `filters` TEXT column, but I could not find a corresponding Flyway migration script. While dev mode uses `drop-and-create`, the `persist` and `prod` profiles require Flyway migrations (`quarkus.flyway.migrate-at-start=true`). Existing production databases will fail with a schema mismatch.

**File:** `core/.../EventSourceEntity.java` line 60

---

### 💡 Suggestion

**4. Pattern compilation on every match is wasteful**

`EventFilterEvaluator.matchesWildcard()` compiles a new `java.util.regex.Pattern` on every invocation. Since this runs per-rule × per-event during each poll cycle (and especially once filter enforcement is added to the pollers), the repeated compilation adds unnecessary overhead.

Consider pre-compiling patterns when the filter configuration is deserialized, e.g. by converting `EventSourceFilterRule` from a simple record to one that lazily caches its compiled pattern.

```java
// Current — compiles on every call
static boolean matchesWildcard(String value, String pattern) {
    ...
    return Pattern.compile(regex.toString()).matcher(value).matches();
}
```

**File:** `core/.../EventFilterEvaluator.java` lines 75–89

---

### 📝 Nitpick

**5. Misleading placeholder text in UI pattern input**

The pattern input field placeholder says `^pattern.*` (line 520 of `EventSourceDetailPage.tsx`), which suggests regex syntax. However, the actual implementation uses **glob patterns** (with `*` and `?` wildcards). The placeholder should reflect this, e.g. `issue-*` or `*[bot]`.

```tsx
<TextInput value={newRulePattern}
    onChange={(_e, v) => setNewRulePattern(v)}
    placeholder="^pattern.*" />  // ← suggests regex, but globs are used
```

**File:** `ui/src/pages/EventSourceDetailPage.tsx` line 520

**6. Updating an event source without filters clears existing filters**

In `applyFields()`, if the incoming `data.getFilters()` is null, the entity's filters are set to null (line 186). This means an update that doesn't include a `filters` field will silently clear any previously configured filters (including the useful default bot/slash-command excludes set during creation). Consider preserving existing filters when the incoming value is null.

**File:** `app/.../EventSourcesResourceImpl.java` lines 178–187

---

### Overall Assessment

The design and architecture of this feature are well-thought-out — the include/exclude rule model with JSON Pointer payload matching and glob patterns is elegant, the dry-run capability is a great UX touch, and the test coverage of `EventFilterEvaluator` is thorough (17 tests covering edge cases).

However, **the PR has a critical gap**: the filters are never actually enforced during event polling, making the feature non-functional for its primary purpose. This needs to be addressed before merging. The dry-run authentication fallback issue (Warning #2) and missing Flyway migration (Warning #3) should also be resolved.

EricWittmann:
Thanks for the thorough review! Here are responses to each item:

---

### 1. Filters are never applied during actual event polling (Critical)

This is actually **by design** — we explicitly chose post-ingest, pre-Manager filtering during the requirements phase. The enforcement point is `PipelineOrchestrator.evaluateFilters()` (called from `processEvent()`), which loads the event source's filters, evaluates them via `EventFilterEvaluator`, and blocks events before the AI Manager LLM call. This was a deliberate trade-off: filtering post-ingest keeps every event in the DB for full auditability, while still preventing the expensive Manager call for filtered events.

The pollers ingest all events → `PipelineOrchestrator` pulls from the queue → `evaluateFilters()` blocks or allows → only allowed events reach the Manager.

### 2. Dry-run endpoint fails for default authentication (Warning)

**Fixed** in `900da22`. `resolveSecretValue()` now takes a `sourceType` parameter and replicates the full poller fallback chain: per-source secret → well-known DB secrets (`GH_TOKEN`/`GITHUB_TOKEN` for GitHub, `JIRA_API_TOKEN` for Jira) → environment variables.

### 3. Missing Flyway migration (Warning)

The migration exists: `app/src/main/resources/db/migration/V34__add_event_source_filters.sql`. It adds the `filters TEXT` column and populates existing rows with default filter rules.

### 4. Pattern compilation on every match (Suggestion)

Agreed this could be optimized, but the current usage is low-frequency — `PipelineOrchestrator` processes events one at a time from the queue, and the dry-run evaluates at most ~100 events per request. Deferring to a follow-up if profiling shows it matters.

### 5. Misleading placeholder text (Nitpick)

**Fixed** in `900da22`. Changed from `^pattern.*` to `*[bot]`.

### 6. Updating without filters clears existing filters (Nitpick)

**Fixed** in `900da22`. `applyFields()` now preserves the entity's existing `filters` value when the incoming data has `null` filters.

PR #168, opened 2026-08-10
feat: label-based action type filtering for Manager (#167)

### LINKED ISSUE #167: Filter action types by label matching against event labels
## Summary

Add label-based filtering of action types so that the Manager only considers action types
relevant to a given event. This enables different action types for different event sources
(or categories of events) based on label matching.

## Motivation

Currently, all action types are passed to the Manager for consideration regardless of the
event source. As the number of action types grows, it becomes important to scope which ones
apply to which events. Labels provide a natural, flexible mechanism for this since event
sources already support labels that propagate to their events.

### Action Type Labels
- Add a `labels` field to action types (same model as event source labels).
- Provide UI for managing labels on the Action Type detail page.

### Filtering Logic
- When passing action types to the Manager for consideration, filter them by label:
  - **No labels on action type** → always included (backwards-compatible default).
  - **Action type has labels** → include only if the action type's labels match the labels
    on the event being considered.
- "Match" means the action type's labels are a subset of (or equal to) the event's labels —
  every label on the action type must be present on the event.

### Affected Components
- **Database / entity model**: add labels column to the action types table.
- **OpenAPI spec**: add `labels` field to the ActionType schema.
- **REST API**: support labels in action type CRUD operations.
- **Manager**: filter the candidate action type list before passing to the AI manager.
- **UI**: add label editing to the Action Type detail page.

EricWittmann:
## Summary

- Add a `labels` field to action types (same model as event source labels) with full CRUD support
  across OpenAPI spec, entity, REST API, import/export, and UI.
- Filter candidate action types in the Manager by label compatibility with the event source:
  action types with no labels are always included (backwards-compatible); action types with labels
  are included only if their labels are a subset of the event's source labels.
- Add unit tests for the label filtering logic covering all matching scenarios.

## Related Issue

Fixes #167

## Test Plan

- [ ] Run `mvn install` to regenerate API beans and verify all tests pass
- [ ] Start the app and navigate to an Action Type detail page — verify the Labels section
      appears in the Info tab
- [ ] Add, edit, and remove labels via the modal — verify they persist after save and page refresh
- [ ] Create two event sources with different labels (e.g. `team-a`, `team-b`)
- [ ] Create action types with labels matching one source but not the other
- [ ] Trigger events from each source and verify the Manager only considers label-compatible
      action types (check activity logs or debug endpoint)
- [ ] Verify action types with no labels are still considered for all events (backwards compat)
- [ ] Test import/export round-trip with action type labels

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/ImportExportService.java
+++ b/app/src/main/java/io/apitomy/axiom/app/ImportExportService.java
@@ -309,6 +309,12 @@ public class ImportExportService {
             entity.maxSteps = item.has("maxSteps") ? item.path("maxSteps").asInt() : null;
             entity.maxBudgetUsd = item.has("maxBudgetUsd") ? item.path("maxBudgetUsd").asDouble() : null;
             entity.environment = jsonOrNull(item, "environment");
+            JsonNode labelsNode = item.path("labels");
+            if (labelsNode.isArray()) {
+                for (JsonNode l : labelsNode) {
+                    entity.labels.add(l.asText());
+                }
+            }
             entity.persist();
             count++;
         }
@@ -438,6 +444,13 @@ public class ImportExportService {
             entity.maxSteps = item.has("maxSteps") ? item.path("maxSteps").asInt() : null;
             entity.maxBudgetUsd = item.has("maxBudgetUsd") ? item.path("maxBudgetUsd").asDouble() : null;
             entity.environment = jsonOrNull(item, "environment");
+            entity.labels.clear();
+            JsonNode labelsNode = item.path("labels");
+            if (labelsNode.isArray()) {
+                for (JsonNode l : labelsNode) {
+                    entity.labels.add(l.asText());
+                }
+            }
             entity.persist();
             if (isNew) created++;
             else updated++;
@@ -600,6 +613,10 @@ public class ImportExportService {
         if (e.maxSteps != null) n.put("maxSteps", e.maxSteps);
         if (e.maxBudgetUsd != null) n.put("maxBudgetUsd", e.maxBudgetUsd);
         putIfNotNull(n, "environment", e.environment);
+        if (e.labels != null && !e.labels.isEmpty()) {
+            var labelsArr = n.putArray("labels");
+            e.labels.forEach(labelsArr::add);
+        }
         return n;
     }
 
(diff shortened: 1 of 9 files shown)


EricWittmann (commented):
## Code Review Summary

This PR adds a `labels` field to action types (mirroring the existing event source labels model) and introduces label-based filtering in the Manager service: action types with no labels are always considered (backwards-compatible), while action types with labels are included only when their labels are a subset of the event source's labels. The change spans the full stack — OpenAPI spec, JPA entity, DB migration, REST API, import/export, Manager filtering logic, unit tests, and UI.

---

### Findings

#### 🟡 Suggestion

**1. `filterByLabels` is not defensive against null `eventSourceLabels`**
`ManagerService.java` — `filterByLabels()`

```java
Set<String> eventLabels = new HashSet<>(eventSourceLabels);
```

If `eventSourceLabels` is ever `null`, this will throw a `NullPointerException`. The caller in `evaluate()` currently always initializes it to `Collections.emptyList()`, so this is safe *today*, but the method is package-visible and could be called from other contexts (including the test class). Adding a null guard would make the method more robust:

```java
Set<String> eventLabels = eventSourceLabels != null
        ? new HashSet<>(eventSourceLabels) : Collections.emptySet();
```

**2. Consider adding a debug log when action types are filtered out by labels**
`ManagerService.java` — `evaluate()`

After filtering action types by labels, there's no log output indicating how many (or which) action types were excluded. Given that this filtering silently reduces the candidate set sent to the AI Manager, a debug-level log would be helpful for troubleshooting:

```java
actionTypes = filterByLabels(actionTypes, eventSourceLabels);
LOG.debugf("Label filter: %d/%d action types remain after filtering",
        actionTypes.size(), totalBeforeFilter);
```

This would help operators diagnose why certain action types aren't being considered.

#### 🟢 Nitpick

**3. Test for null `eventSourceLabels` parameter is missing**
`ManagerServiceLabelFilterTest.java`

The tests cover null action type lists and null labels on individual action types, but there's no test for passing `null` as the `eventSourceLabels` parameter. This would document the expected behavior and guard against the issue in finding #1.

---

### Overall Assessment

**✅ Looks good to merge.** This is a clean, well-structured PR that follows existing patterns closely — the entity mapping, migration, import/export handling, and UI integration all mirror the established event source labels implementation. The `filterByLabels` logic is correct and well-tested with 10 unit tests covering all meaningful scenarios. The backwards-compatibility guarantee (label-less action types always included) is a good design choice. The suggestions above are minor robustness improvements, not blockers.

PR #174, opened 2026-08-10
fix: DiskUsageWidget Top-N correctness — sort by usage descending

### LINKED ISSUE #169: DiskUsageWidget: Top-N correctness depends on server-side ordering
## Context

Raised during review of PR #163 (Disk Usage Breakdown widget).

## Problem

`DiskUsageWidget` calls `fetchDiskUsage(1, maxProjects)` to fetch a single page of results, then
sorts client-side by `diskUsageBytes` descending. If the backend doesn't guarantee results are
ordered by usage, higher-usage projects that fall on later pages will be missed — the chart would
show the wrong "top N" projects.

## Suggested Fix

Confirm whether the `/usage/disk` endpoint sorts by usage descending. If it doesn't, either:

1. Add server-side sorting by `diskUsageBytes` descending, or
2. Fetch a larger page client-side and slice to `maxProjects` (similar to how
   `AiCostByProjectWidget` fetches up to 50 projects then slices)

EricWittmann:
## Summary

Fixes #169

The `DiskUsageWidget` was calling `fetchDiskUsage(1, maxProjects)` and sorting client-side by `diskUsageBytes`, but the server-side `/usage/disk` endpoint was sorting results alphabetically by project name. This meant that projects with the highest disk usage could be missed if they fell on later pages — the widget would show the wrong "top N" projects.

### Server-side (`UsageResourceImpl.java`)
- Changed the disk usage query sort order from `Sort.ascending("name")` to `Sort.descending("diskUsageBytes")`, ensuring the highest-usage projects are always returned first.

### Client-side (`DiskUsageWidget.tsx`)
- Changed `fetchDiskUsage(1, maxProjects)` to `fetchDiskUsage(1, 50)` to fetch a larger pool of projects (matching the pattern used in `AiCostByProjectWidget`).
- Added `.slice(0, maxProjects)` after sorting to trim results to the configured maximum.

This two-pronged approach ensures correctness: the server returns highest-usage projects first, and the client fetches a generous page then slices to the desired top-N count.

### DIFF
--- a/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
@@ -145,7 +145,7 @@ public class UsageResourceImpl implements UsageResource {
         long totalCount = ProjectEntity.count(hql.toString(), params);
 
         List<DiskUsageProject> items = ProjectEntity.<ProjectEntity>find(
-                        hql.toString(), Sort.ascending("name"), params)
+                        hql.toString(), Sort.descending("diskUsageBytes"), params)
                 .page(Page.of(pageNum - 1, pageSize))
                 .list()
                 .stream()
--- a/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
+++ b/app/src/main/java/io/apitomy/axiom/app/rest/UsageResourceImpl.java
@@ -7,6 +7,7 @@ import io.apitomy.axiom.api.beans.DiskUsageProject;
 import io.apitomy.axiom.api.beans.DiskUsageSearchResults;
 import io.apitomy.axiom.core.entities.AiUsageEntity;
 import io.apitomy.axiom.core.entities.ProjectEntity;
+import io.quarkus.hibernate.orm.panache.PanacheQuery;
 import io.quarkus.panache.common.Page;
 import io.quarkus.panache.common.Sort;
 import io.smallrye.common.annotation.RunOnVirtualThread;
@@ -130,7 +131,8 @@ public class UsageResourceImpl implements UsageResource {
      */
     @Override
     public DiskUsageSearchResults listDiskUsage(BigInteger page, BigInteger limit,
-                                                 String filterName) {
+                                                 String filterName, String sortBy,
+                                                 String sortOrder) {
         int pageNum = page != null ? page.intValue() : 1;
         int pageSize = limit != null ? limit.intValue() : 20;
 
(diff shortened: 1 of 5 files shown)


EricWittmann (commented):
## Code Review Summary

This PR fixes the `DiskUsageWidget` top-N correctness issue by: (1) changing the server-side disk usage query sort from alphabetical (`name` ascending) to `diskUsageBytes` descending, and (2) fetching a larger page of 50 items on the client side and slicing to `maxProjects` after filtering. The approach is sound and matches the established pattern in `AiCostByProjectWidget`.

---

### ⚠️ Warning

**`DiskUsagePage` sort order also changes (side effect)**
`UsageResourceImpl.java` — The sort order change from `Sort.ascending("name")` to `Sort.descending("diskUsageBytes")` applies globally to the `/usage/disk` endpoint. This also changes the default sort order on the `DiskUsagePage`, which renders a paginated table of all projects' disk usage. Previously rows were sorted alphabetically by project name; now they'll be sorted by disk usage descending.

This is arguably a *better* default for a disk usage table, but it's an undocumented behavioral change that affects a page beyond the widget. Worth confirming this is intentional.

---

**NULL ordering for `diskUsageBytes` in descending sort**
`ProjectEntity.java` — The `diskUsageBytes` field is `Long` (nullable) with no default column value. In PostgreSQL (typical Quarkus production DB), `ORDER BY diskUsageBytes DESC` places `NULL` values **first**. If there are many projects that have never had their disk usage computed (null values), they would fill the first page of results, be mapped to `0` by `toDiskUsageBean`, filtered out by `.filter(p => p.bytes > 0)` on the client, and the widget could show fewer results than expected — or none at all.

Consider using `COALESCE` in the HQL sort expression or adding `NULLS LAST` to the ordering to ensure projects with actual disk usage always sort first. For example, switching to a raw HQL `ORDER BY` clause:
```java
hql.append(" ORDER BY COALESCE(diskUsageBytes, 0) DESC");
```
and dropping the `Sort` parameter. Alternatively, this could be addressed in a follow-up if the current data model guarantees non-null values in practice.

---

### 💡 Suggestion

**Client-side sort is now redundant**
`DiskUsageWidget.tsx` (line 27) — Since the server now returns results sorted by `diskUsageBytes` descending, the client-side `.sort((a, b) => b.bytes - a.bytes)` is redundant. It's not harmful (defensive coding is fine), but if you want to keep the code minimal, the sort could be removed. Keeping it is also reasonable as a safety net.

---

### 🔍 Nitpick

**Magic number `50`**
`DiskUsageWidget.tsx` (line 21) — The hardcoded `50` could be extracted as a named constant (e.g., `const FETCH_POOL_SIZE = 50`) for clarity, matching how `maxProjects` is already a named value. The same applies to `AiCostByProjectWidget` which uses the same pattern — but that's outside the scope of this PR.

---

### Overall Assessment

**Looks good to merge** with one consideration: the NULL ordering concern is worth evaluating based on your production data. If `diskUsageBytes` is reliably populated for all projects, it's a non-issue. If it can be null for many projects, a `COALESCE` or `NULLS LAST` would make the fix more robust. The `DiskUsagePage` sort order change is arguably an improvement, just worth acknowledging.

Clean, well-scoped fix. 👍

EricWittmann:
Decision:  Sort by and sort direction should be added to the REST API so that the client can choose how to sort the results.

PR #182, opened 2026-08-12
Add refresh button to Dashboard page

### LINKED ISSUE #177: Add refresh button to Dashboard page
## Description

The Dashboard page currently has no way to refresh the contents of all widgets without reloading
the entire browser page. A refresh button should be added to allow users to re-fetch data for all
widgets at once.

## Acceptance Criteria

- A refresh button is visible on the Dashboard page (consistent with the refresh buttons on other
  list pages like Tools, Action Types, etc.)
- Clicking the button triggers all widgets on the dashboard to reload their data
- The button should use the same `SyncAltIcon` pattern used elsewhere in the app

EricWittmann:
## Summary

Adds a refresh button to the Dashboard view page that triggers all widgets to reload their data without a full browser page reload.

## Changes

- Added a refresh button (using `SyncAltIcon`) to the Dashboard view page toolbar, visible when not in edit mode
- The button uses a React key-based remount strategy: clicking it increments a `refreshKey` counter that is applied as a `key` prop on each `WidgetCard`, causing React to unmount and remount all widget components — which in turn re-triggers their `useEffect` data-fetching hooks
- Button style (`variant="plain"`, `aria-label="Refresh"`) is consistent with refresh buttons on other pages like ProjectDetailPage

## Files Modified

- `ui/src/pages/DashboardViewPage.tsx`

Fixes #177

### DIFF
--- a/ui/src/pages/DashboardViewPage.tsx
+++ b/ui/src/pages/DashboardViewPage.tsx
@@ -23,6 +23,7 @@ import PencilAltIcon from "@patternfly/react-icons/dist/esm/icons/pencil-alt-ico
 import PlusCircleIcon from "@patternfly/react-icons/dist/esm/icons/plus-circle-icon";
 import SaveIcon from "@patternfly/react-icons/dist/esm/icons/save-icon";
 import StarIcon from "@patternfly/react-icons/dist/esm/icons/star-icon";
+import SyncAltIcon from "@patternfly/react-icons/dist/esm/icons/sync-alt-icon";
 import TimesIcon from "@patternfly/react-icons/dist/esm/icons/times-icon";
 import TrashIcon from "@patternfly/react-icons/dist/esm/icons/trash-icon";
 import {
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review Summary

This PR adds a refresh button to the Dashboard view page that uses a React key-based remount strategy to force all `WidgetCard` components to unmount and remount, re-triggering their data-fetching `useEffect` hooks.

## Analysis

**Approach:** The `refreshKey` counter pattern is a well-established React idiom for forcing component remounts. Each `WidgetCard` sits inside a `<div key={w.id}>` (used by the grid layout for positioning), and the inner `key={refreshKey}` on `WidgetCard` correctly causes React to unmount/remount the widget when the key value changes — since each `WidgetCard` is the sole keyed child within its parent `<div>`.

**Consistency:** The button styling (`variant="plain"`, `aria-label="Refresh"`, `SyncAltIcon`) matches the existing refresh button pattern in `ProjectDetailPage.tsx` (lines 218–225). The icon import follows the same `@patternfly/react-icons/dist/esm/icons/` convention used by the other imports in this file.

**Correctness:** The refresh button is only rendered in view mode (not edit mode), so `refreshKey` won't change while editing. The `key={refreshKey}` is applied to all `WidgetCard` instances regardless of mode, but since the counter only increments from the view-mode button, this is harmless.

## Findings

No issues found. This is a clean, focused change that follows existing codebase patterns.

## Verdict

✅ **Looks good to merge.** Small, well-scoped change with a sound approach.

PR #184, opened 2026-08-12
Scheduled Jobs: CRON-style automation for Axiom (#171)

### LINKED ISSUE #171: Scheduled Jobs: CRON-style automation for Axiom
## Summary

Add a first-class **Scheduled Jobs** feature to Axiom, enabling CRON-style automation that runs
on a configurable schedule. Scheduled jobs are global (not tied to projects), define their own
execution (prompt template or bash script), and can perform any work — emit events, create
projects, interact with multiple repos, call Axiom APIs, etc.

## Motivation

Axiom currently reacts to external events (GitHub, Jira) and triages them through the AI Manager.
There is no way to run work on a schedule — things like nightly dependency audits, weekly summary
reports across repos, periodic stale-issue cleanup, or scheduled health checks. The existing
`ReportScheduler` proves the pattern works but is tightly coupled to report generation. A
general-purpose scheduled jobs feature would unlock a broad class of automation use cases.

### Core Model

A new top-level entity: `ScheduledJobEntity`. Key properties:

- **Identity:** `name`, `description`, `labels`, `enabled` (toggle to pause without deleting)
- **Schedule:** Simple presets consistent with existing report scheduling:
  - `schedulePreset`: `hourly`, `daily`, `weekly`, `monthly`, `none`
  - `scheduleTime`: time of day (e.g., `"08:00"`)
  - `scheduleDayOfWeek`: day of week for weekly preset
  - `nextRunAt` / `lastRunAt`: computed timestamps for the scheduler
- **Execution:** Each job defines its own execution, independent of action types:
  - `executionMode`: `actor` (AI) or `script` (bash)
  - `promptTemplate`: for actor mode — supports template variables
  - `scriptTemplate`: for script mode — supports template variables
  - `model`: optional AI model override (actor mode)
  - `engine`: optional AI engine override (actor mode)
  - `allowedTools`: tools available to the actor (actor mode)
  - `maxSteps` / `maxBudgetUsd`: execution limits
  - `environment`: optional JSON object of environment variables (supports `${secret:NAME}`)

### Execution Behavior

- **Global scope:** Scheduled jobs are not tied to any project. They run in their own context and
  can create projects, emit events, or interact with any part of the system.
- **Actor mode:** Launches an AI actor with the job's prompt template, model, engine, and allowed
  tools. Runs without a project/task context.
- **Script mode:** Runs the job's bash script template via the existing script execution
  infrastructure. Template variables like `{{apiBaseUrl}}`, `{{jobName}}`, etc. are available.
- **Manual trigger:** A "Run Now" action that immediately executes the job outside its normal
  schedule, for testing and one-off needs.

### Run History & Observability

- **Dedicated run history:** A new `ScheduledJobRunEntity` to record each execution:
  - `jobId`, `status` (pending/running/completed/failed), `startedAt`, `completedAt`, `output`,
    `error`
  - Visible from the job's detail page as a run history table
- **Activity log integration:** Each run also creates entries in the existing activity log for
  global visibility
- **SSE events:** Fire `SseEvent` CDI events for real-time UI updates when jobs start/complete

### Scheduler

- A new `ScheduledJobPoller` (similar to `ReportScheduler`):
  - `@Scheduled` method polling every 60s (configurable)
  - Finds jobs where `enabled = true` and `nextRunAt <= now()`
  - Dispatches execution and updates `nextRunAt` / `lastRunAt`
  - Uses `concurrentExecution = SKIP` to prevent overlap

### REST API (Contract-First)

New endpoints in the OpenAPI spec (`common/api/src/main/resources/openapi.json`):

- `GET /api/v1/scheduled-jobs` — List all scheduled jobs
- `POST /api/v1/scheduled-jobs` — Create a scheduled job
- `GET /api/v1/scheduled-jobs/{id}` — Get a scheduled job
- `PUT /api/v1/scheduled-jobs/{id}` — Update a scheduled job
- `DELETE /api/v1/scheduled-jobs/{id}` — Delete a scheduled job
- `POST /api/v1/scheduled-jobs/{id}/run` — Manually trigger a run
- `GET /api/v1/scheduled-jobs/{id}/runs` — List run history for a job

### UI

- New **"Scheduled Jobs"** page in the sidebar (Automation section, alongside Event Sources and
  Action Types)
- **List page:** Table of all scheduled jobs with name, schedule, last run, next run, status,
  enabled toggle
- **Detail page:** Job configuration form + run history table with status, start time, duration,
  and output/error
- **Run Now** button on the detail page

## Design Notes

- Follow the existing `ReportDefinitionEntity` + `ReportScheduler` pattern for schedule
  management (`nextRunAt`/`lastRunAt` tracking, `computeNextRunAt()` logic)
- Mirror `ActionTypeEntity` field patterns for the execution definition (prompt/script templates,
  model, engine, allowed tools, environment)
- Use database-driven configuration (Panache entity), consistent with all other Axiom config
- Contract-first: define OpenAPI spec first, generate JAX-RS interfaces, then implement
- Labels support for organization and filtering

## Out of Scope (for now)

- Full cron expression syntax (presets cover 90% of use cases)
- Project-scoped scheduled jobs
- Chaining / dependencies between scheduled jobs
- Concurrency controls (running multiple instances of the same job in parallel)

EricWittmann:
## Summary

- Add a new top-level **Scheduled Jobs** entity with CRON-style scheduling (hourly/daily/weekly/monthly),
  dual execution modes (AI actor via AiEngine, bash script via ProcessBuilder), run history tracking,
  and manual trigger support
- Full-stack implementation: OpenAPI contract (5 paths, 8 operations, 4 schemas), Flyway migration,
  Panache entities, validator, scheduler/queue consumer/execution service, REST resource, import/export,
  SSE events, and React UI (list page, detail page with run history)
- Integrated with Configuration Assistant (system prompt, MCP tools, validation, apply flow),
  Configuration Packs (import/export UI), and documentation (7 doc files updated)

## Related Issue

Fixes #171

## Test Plan

- [ ] Run `./build.sh` — all 319 tests pass, including 37 new `ScheduledJobValidatorTest` tests
- [ ] Run `npm run build` in `ui/` — TypeScript compiles cleanly
- [ ] Start the app (`./dev.sh`) and verify the V36 migration runs without errors
- [ ] Navigate to Configuration > Scheduled Jobs in the sidebar
- [ ] Create a scheduled job (actor mode, daily schedule) and verify it appears in the list
- [ ] Edit the job on the detail page — verify save, enable/disable toggle, and delete work
- [ ] Click "Run Now" and verify a run appears in the run history table
- [ ] Create a script-mode job with `echo "hello"` and trigger it manually
- [ ] Test import/export: export a config pack including a scheduled job, then re-import it
- [ ] Open the Configuration Assistant and ask it to create a scheduled job — verify it generates
  valid JSON in the `scheduled-jobs/` directory and applies successfully


EricWittmann (commented):
## Code Review: Scheduled Jobs — CRON-style Automation

This PR adds a full-stack "Scheduled Jobs" feature, enabling users to define recurring or ad-hoc jobs that execute either via an AI agent (actor mode) or a bash script (script mode). The implementation spans the OpenAPI spec, JPA entities, a Quarkus scheduler/queue system, REST API, validation, configuration pack import/export, SSE events, and a complete React frontend with list and detail pages.

This is a very well-structured feature that follows the existing project patterns closely. The code is clean, well-documented, and integrates into every relevant system (packs, SSE, tracing, activity log, AI usage tracking). Below are my findings.

---

### 🔴 Critical

**1. Script template injection via job name**
`ScheduledJobExecutionService.java` — `resolveScriptTemplate()`

The `{{jobName}}` placeholder is substituted directly into a bash script with no escaping:
```java
return template
    .replace("{{jobName}}", job.name != null ? job.name : "")
    ...
```
Since `job.name` comes from user input, a name containing characters like `` `; rm -rf /` `` or `$(malicious-command)` will be interpreted by bash. The same issue applies to `resolvePromptTemplate()` but with lower risk since the prompt goes to an AI agent, not a shell.

**Recommendation:** Shell-escape the job name before substitution, or at minimum validate that job names only contain safe characters (the validator does not currently check this).

**2. `executeScript()` can block indefinitely on large output**
`ScheduledJobExecutionService.java:214`

```java
Process process = pb.start();
String output = new String(process.getInputStream().readAllBytes());
boolean finished = process.waitFor(scriptTimeoutSeconds, TimeUnit.SECONDS);
```

`readAllBytes()` blocks until the stream is closed (i.e., the process exits or the pipe buffer fills). If the script produces unbounded output, this can block the consumer thread forever — the `waitFor` timeout never executes. Additionally, `redirectErrorStream(true)` means stderr is merged in, doubling the risk.

**Recommendation:** Read the stream in a separate thread or use a bounded buffer, and enforce the timeout around the full operation, not just `waitFor`.

---

### 🟡 Warning

**3. Monthly schedule skips the current day**
`ScheduledJobScheduler.java:149`

```java
case "monthly" -> {
    return todayAtTime.plusMonths(1).toInstant();
}
```

Unlike the `daily` case which checks `if (todayAtTime.isAfter(now))`, the `monthly` case unconditionally adds a month. If a user creates a monthly job at 7:00 AM for 8:00 AM today, it won't run until next month. The same applies to `computeInitialNextRunAt()` since it delegates to the same method.

**4. Weekly schedule with `TemporalAdjusters.next()` always skips to next week**
`ScheduledJobScheduler.java:142`

```java
ZonedDateTime nextOccurrence = todayAtTime.with(TemporalAdjusters.next(targetDay));
```

`TemporalAdjusters.next()` always advances to the *next* occurrence, even if today is the target day and the time hasn't passed yet. For a weekly Monday job being created on Monday morning, this will skip to next Monday. Consider using `nextOrSame()` with a time check.

**5. Stale "Running" runs not recovered on startup**
`ScheduledJobQueueConsumer.java:55-61`

```java
ScheduledJobRunEntity.<ScheduledJobRunEntity>list(
    "status = 'Pending' order by createdOn asc")
    .forEach(r -> { ... queue.add(r.id); });
```

Only "Pending" runs are re-enqueued on startup. If the application crashes while a job is "Running", those runs will be stuck in "Running" status permanently with no way to recover without manual database intervention.

**Recommendation:** Also find runs with `status = 'Running'` on startup and either mark them as "Failed" (with a note that they were interrupted) or re-enqueue them.

**6. `createScheduledJob` returns HTTP 200 instead of 201**
`openapi.json:4531`

The OpenAPI spec defines `createScheduledJob` with a `200` response, while the convention used by other creation endpoints in this API (`createProject`, `createActionType`, `createEventSource`, etc.) is `201`. This inconsistency could confuse API consumers.

---

### 💡 Suggestion

**7. Detail page doesn't subscribe to SSE for live run status updates**
`ScheduledJobDetailPage.tsx`

The backend fires `scheduled-job-run-updated` SSE events when runs change status, but the detail page doesn't subscribe to them. Users need to manually reload the page to see run progress. Other pages in the app (e.g., inbox) already subscribe to SSE events for live updates. Adding an SSE subscription that refreshes the runs list when a `scheduled-job-run-updated` event arrives would improve the user experience.

**8. `getKnownNames()` loads all secrets, tools, and toolsets**
`ScheduledJobsResourceImpl.java:211-236`

The method loads all `SecretEntity`, `ToolDefinitionEntity`, and `ToolsetEntity` records into memory to build the known-names set for validation. This is fine for current scale but could be optimized with targeted queries (e.g., `SELECT name FROM ...`) if the entity counts grow large.

**9. Error feedback missing in create modal**
`ScheduledJobsPage.tsx:74-79`

The `handleCreate` function swallows errors with `console.error`. If creation fails (e.g., duplicate name → 422), the user sees no feedback. Consider adding a toast notification or error banner.

---

### 📝 Nitpick

**10. Consumer thread uses `Thread` instead of virtual threads**
`ScheduledJobQueueConsumer.java:63`

The rest of the application uses `@RunOnVirtualThread`. The consumer thread is a daemon platform thread. While this works (and the thread is long-lived so virtual threads wouldn't help much), it's a minor inconsistency.

**11. `AssistantApplyResult` schema includes scheduledJobs counts but not eventSources counts**
`openapi.json:6987-6998`

The `AssistantApplyResult` schema includes `scheduledJobs`, `scheduledJobsCreated`, `scheduledJobsUpdated` (good!) but I notice the existing `eventSources` created/updated fields are missing from the schema even though `UpsertResult` in the backend tracks them. This is a pre-existing issue, not introduced by this PR.

---

### Overall Assessment

**✅ Looks good to merge with minor changes recommended.** This is a thorough, well-implemented feature that follows the project's established patterns consistently across the full stack. The critical items (script injection and the unbounded process output read) should be addressed, and the schedule computation edge cases in weekly/monthly are worth fixing. The remaining items are lower priority and can be tracked as follow-up issues.

PR #302, opened 2026-09-03
feat: add UNIQUE constraint and index to scheduled_job_label table

### LINKED ISSUE #240: Fix: scheduled_job_label missing UNIQUE constraint and index
## Problem

The `scheduled_job_label` join table is missing a `UNIQUE(scheduled_job_id, label)` constraint and
an explicit index, unlike every other `*_label` join table in the schema:

| Table | Has UNIQUE constraint | Has explicit index |
|-------|-----------------------|--------------------|
| `tool_label` | Yes | `idx_tool_label(label, tool_id)` |
| `project_label` | Yes | `idx_project_label(label, project_id)` |
| `report_label` | Yes | `idx_report_label(label, report_id)` |
| `report_definition_initial_label` | Yes | `idx_report_definition_initial_label(label, report_definition_id)` |
| `event_source_label` | Yes | `idx_event_source_label(label, event_source_id)` |
| `dashboard_label` | Yes | `idx_dashboard_label(label, dashboard_id)` |
| `action_type_label` | Yes | `idx_action_type_label(label, action_type_id)` |
| **`scheduled_job_label`** | **No** | **No** |

This is likely an oversight from when the table was added. Without the constraint, duplicate labels
can be inserted for the same scheduled job.

## Proposed Fix

Add a Flyway migration with:

1. `UNIQUE(scheduled_job_id, label)` constraint
2. `idx_scheduled_job_label(label, scheduled_job_id)` index (matching the pattern of other label
   tables)

matheusandre1:
Closes: #240

### DIFF
--- a/app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
+++ b/app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
@@ -1,0 +1,20 @@
+-- scheduled_job_label was created (V36) without the UNIQUE constraint and index
+-- that every other *_label join table has. Bring it in line with the rest of the schema.
+
+-- Remove any duplicate (scheduled_job_id, label) rows that may already exist so the
+-- UNIQUE constraint below can be added safely. The table has no primary key, so
+-- rebuild its contents from a distinct copy held in a session-scoped temporary table.
+-- H2 auto-commits DDL, so a temp table guarantees a repaired re-run starts clean.
+CREATE LOCAL TEMPORARY TABLE scheduled_job_label_dedup AS
+    SELECT DISTINCT scheduled_job_id, label FROM scheduled_job_label;
+
+DELETE FROM scheduled_job_label;
+
+INSERT INTO scheduled_job_label (scheduled_job_id, label)
+    SELECT scheduled_job_id, label FROM scheduled_job_label_dedup;
+
+DROP TABLE scheduled_job_label_dedup;
+
+ALTER TABLE scheduled_job_label ADD CONSTRAINT IF NOT EXISTS uq_scheduled_job_label UNIQUE (scheduled_job_id, label);
+
+CREATE INDEX IF NOT EXISTS idx_scheduled_job_label ON scheduled_job_label(label, scheduled_job_id);
--- a/app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
+++ b/app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
@@ -2,18 +2,15 @@
 -- that every other *_label join table has. Bring it in line with the rest of the schema.
 
 -- Remove any duplicate (scheduled_job_id, label) rows that may already exist so the
--- UNIQUE constraint below can be added safely. The table has no primary key, so
--- rebuild its contents from a distinct copy held in a session-scoped temporary table.
--- H2 auto-commits DDL, so a temp table guarantees a repaired re-run starts clean.
-CREATE LOCAL TEMPORARY TABLE scheduled_job_label_dedup AS
-    SELECT DISTINCT scheduled_job_id, label FROM scheduled_job_label;
-
-DELETE FROM scheduled_job_label;
-
-INSERT INTO scheduled_job_label (scheduled_job_id, label)
-    SELECT scheduled_job_id, label FROM scheduled_job_label_dedup;
-
-DROP TABLE scheduled_job_label_dedup;
+-- UNIQUE constraint below can be added safely. The table has no primary key, so use
+-- H2's built-in _ROWID_ pseudo-column to keep the lowest-rowid row of each group and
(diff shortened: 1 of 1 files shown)


app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
```diff
@@ -0,0 +1,19 @@
...
+-- rebuild its contents from a distinct copy (portable across H2 and PostgreSQL).
+CREATE TABLE scheduled_job_label_dedup AS
+    SELECT DISTINCT scheduled_job_id, label FROM scheduled_job_label;
+
+DELETE FROM scheduled_job_label;
+
+INSERT INTO scheduled_job_label (scheduled_job_id, label)
+    SELECT scheduled_job_id, label FROM scheduled_job_label_dedup;
+
+DROP TABLE scheduled_job_label_dedup;
```
EricWittmann: **Non-atomic + non-re-runnable dedup under H2.** Because H2 auto-commits each DDL statement, `CREATE TABLE` (line 7) and `DROP TABLE` (line 15) each commit immediately, along with the intervening `DELETE`/`INSERT`. Two consequences:

1. If a later statement fails (e.g. `ADD CONSTRAINT` on line 17), the delete-and-rebuild has already been committed with **no** unique constraint in place — the migration is marked failed but the data change is not rolled back.
2. If any statement here fails, the auto-committed `scheduled_job_label_dedup` table is left behind. A re-run (after `flyway repair`) then fails at line 7 with *"table already exists"* — permanently stuck without a manual `DROP`. This violates the repo's "migrations must be idempotent" rule.

True atomicity isn't achievable under H2 DDL auto-commit, but using a **session-scoped temporary table** makes a repaired re-run start clean (temp tables don't persist across Flyway sessions):

```suggestion
CREATE LOCAL TEMPORARY TABLE scheduled_job_label_dedup AS
    SELECT DISTINCT scheduled_job_id, label FROM scheduled_job_label;

DELETE FROM scheduled_job_label;

INSERT INTO scheduled_job_label (scheduled_job_id, label)
    SELECT scheduled_job_id, label FROM scheduled_job_label_dedup;

DROP TABLE scheduled_job_label_dedup;
```

app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
```diff
@@ -0,0 +1,19 @@
...
+
+ALTER TABLE scheduled_job_label ADD CONSTRAINT uq_scheduled_job_label UNIQUE (scheduled_job_id, label);
```
EricWittmann: **Inconsistent idempotency.** The index on line 19 uses `CREATE INDEX IF NOT EXISTS`, but this `ADD CONSTRAINT` has no guard. Since constraint DDL auto-commits in H2, a repaired re-run after the constraint was already added fails with *"constraint already exists"*. H2 supports `IF NOT EXISTS` here:

```suggestion
ALTER TABLE scheduled_job_label ADD CONSTRAINT IF NOT EXISTS uq_scheduled_job_label UNIQUE (scheduled_job_id, label);
```

app/src/main/resources/db/migration/V56__add_scheduled_job_label_unique_and_index.sql
```diff
@@ -0,0 +1,19 @@
+-- scheduled_job_label was created (V36) without the UNIQUE constraint and index
+-- that every other *_label join table has. Bring it in line with the rest of the schema.
+
+-- Remove any duplicate (scheduled_job_id, label) rows that may already exist so the
+-- UNIQUE constraint below can be added safely. The table has no primary key, so
+-- rebuild its contents from a distinct copy (portable across H2 and PostgreSQL).
```
EricWittmann: **Misleading comment.** This project is H2-only (`quarkus.datasource.db-kind=h2`), so "portable across H2 and PostgreSQL" doesn't apply and may steer future maintainers away from H2-native constructs (e.g. temporary tables). Suggest:

```suggestion
-- rebuild its contents from a distinct copy. (The datasource is H2-only.)
```

EricWittmann (commented):
Thanks for bringing `scheduled_job_label` in line with the other `*_label` tables. The happy path is correct (dedup → UNIQUE + index, matching the `(label, id)` index convention). My comments are all about **robustness under H2's DDL auto-commit semantics** — this project is H2-only, and H2 commits each DDL statement immediately, so a partial failure can leave the DB in a committed, inconsistent, non-re-runnable state. Inline suggestions below.

Note: the in-memory dev datasource always starts empty (dedup is a no-op there), so the real exposure is the file-based `%prod`/`%persist` H2 databases.

EricWittmann (approved):
Reviewed and fixed the migration dedup logic. Replaced the temp-table rebuild with a targeted `_ROWID_`-keyed delete that removes only actual duplicates, touches no rows in the common case, and never empties the table (eliminating the interruption data-loss window). Thanks @matheusandre1 for the original contribution.

EricWittmann:
Thanks @matheusandre1 :+1:

